PR #32, opened 2018-04-17
Redesign sawtooth adapter to make it easy to add new use uses

Ram-srini:
1. Generalize the sawtooth adapter
2. Move use case specific code like batch creation and adddress creation to application adapter
3. Each application has to implement batch builder interface.

Signed-off-by: Ramakrishna Srinivasamurthy <ramakrishna.srinivasamurthy@intel.com>

### DIFF
--- a/src/sawtooth/Application/BatchBuilder.js
+++ b/src/sawtooth/Application/BatchBuilder.js
@@ -1,0 +1,17 @@
+'use strict'
+
+class BatchBuilder {
+    constructor() {
+    }
+
+    buildBatch(args) {
+        throw new Error('buildiBatch is not implemented for this application!!');
+    }
+
+    calculateAddress(name) {
+        throw new Error('calculateAddress is not implemented for this application!!');
+    }
+
+}
+
+module.exports = BatchBuilder;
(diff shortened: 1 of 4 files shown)


Ram-srini:
@feihujiang @haojun I have made changes sawtooth adapter to have a application independent code and moved application specific logic to application adapter. Please review the changes. I will raise one more PR for small bank use this week.

haojun:
Hi @Ram-srini,
One of my concern is that ideally the adapter should be independent of use cases, and use the configuration file to fit various use case. But according to your new design, developers have to extend the implementation of adapter (add new builder) for each new use case, that may be not a good way.

A simple idea is to implement builder as external script and specify it in the configuration file, so an adapter can invoke the use case specified batch builder. 

BTW, I'm not a sawtooth expert, I just wonder why it is not feasible to have a common batch builder for all use cases?

Ram-srini:
@haojun Yes sawtooth adapter should be independent of use case, But application can different data format to store eg. json, protobug, csv etc. For that use case sawtooth adapter shouldn't be changed. Also address calculation is also dependent on application. So i moved those 2 parts payload construction and address  calculation from sawtooth adapter and created to application adapter. Writing new bench mark use case user has to write the logic build payload data and address calculation logic.

PR #52, opened 2018-04-30
Simplify install document

davidkhala:
`npm install` of fabric-sdk-node can fits for most cases.
And seemingly there is some duplicated documents in benchmark/simple/main


davidkhala:
noted，will update later

PR #58, opened 2018-05-04
add specific version of grpc in package.json

### LINKED ISSUE #56: Fabric eventhub error
It seems that the new version of grpc (`v1.11.2`, which will be installed when `npm install fabric-client`) will cause a EventHub error when joining channel

haojun:
closes #56 


package.json
```diff
@@ -24,6 +24,7 @@
     "commander": "^2.11.0",
     "dockerode": "^2.5.0",
     "fs-extra": "^4.0.2",
+    "grpc": "^1.10.1",
```
nklincoln: Would recommend removing the hat for now and use:

`"grpc": "1.10.1",`

PR #66, opened 2018-05-09
Support batch in framework

haojun:
1. Update blockchain interfaces and txns count calculation to support batch
2. Update sawtooth code to follow linting rules

### DIFF
--- a/src/sawtooth/Application/SimpleBatchBuilder.js
+++ b/src/sawtooth/Application/SimpleBatchBuilder.js
@@ -3,22 +3,34 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  *
- * @file, batchBuilder for simple use case and it use case specific logic to
- * buildBatch and calculateAddress
  **/
 
-'use strict'
+'use strict';
 
-var BatchBuilder = require('./BatchBuilder.js')
+const BatchBuilder = require('./BatchBuilder.js');
 
+/**
+ * BatchBuilder for simple use case and it use case specific logic to
+ * buildBatch and calculateAddress
+ */
 class SimpleBatchBuilder extends BatchBuilder {
 
+    /**
+     * Constructor
+     * @param {String} fName transaction family name
+     * @param {String} fVersion transaction family version
+     */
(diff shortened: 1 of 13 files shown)


haojun:
@Ram-srini @feihujiang 
I've updated the code to correct the method of calculating txns count for batch. 
Please note now in case of batch, the txNumb as well as tps in configuration file are also used to represent txn's count (rather than batch's count) as in no-batch case. 
I also update the Sawtooth code to obey eslint rules.

Ram-srini:
@haojun 
Is it requires any changes to simple benchmark config file? As it is not having multiple transactions in a batch. Is simple benchmark use case works as expected with you changes?

haojun:
@Ram-srini 
Nothing to change for simple benchmark, the current code is compatible with no-batch case

Ram-srini:
@haojun if we specify txNumber and tps in terms of transactions then how caliper will submit the transaction rate? For example suppose txNumber=1000, tps=100 and if i choose txnPerBatch=10, will caliper submits at a rate 10batches/sec? And what if change txnPerBatch=20, whether rate change to 5batches/sec?

haojun:
@Ram-srini yes, the rate is 10 and 5 batches/sec for your cases. Actually, rate controllers have no idea about batch, they only control the speed according to the configured value as well as the number of txns that have already been submitted. 
You can find a submitCallback which is added in the context and used to count the number of txns once new txns are submitted (in case of batch, multiple txns will be counted).

src/sawtooth/Application/SimpleBatchBuilder.js
```diff
@@ -80,7 +97,6 @@ class SimpleBatchBuilder extends BatchBuilder {
         const familyNameSpace = _hash(this.familyName).substring(0, 6);
         let address = familyNameSpace + _hash(name).slice(-64);
         return address;
-    }
- 
+    }
```
feihujiang: Has a lint error

haojun:
@feihujiang  Fixed

PR #71, opened 2018-05-18
support smallbank with fabric

haojun:
Main updates:
1. Add chaincode of smallbank
2. Update fabric adaptor to support submit multiple txns using invokesmartcontract()
3. Minor changes for smallbankOperations.js, removing the test of amalgamate txn (Amalgamating causes subsequent operations on this account to fail, which is not suitable from the point of performance test)
4. Add fabric related configuration files for smallbank test
5. Add a Readme doc in the smallbank folder

add small bank config for fabric

Bugfix and add readme for smallbank benchmark


haojun:
@Ram-srini @feihujiang , this is the PR for smallbank with fabric

Ram-srini:
@haojun Why did you remove amalgamate operation? What is the failure you are getting with amalgamate operation. We can fix that issue. I didn't see any issue with sawtooth, is it failing only on fabric?

Ram-srini:
@haojun I ran the benchmark test with your chain code of small bank. I got the many errors (500/1500) in the transactions, error: [client-utils.js]: sendPeersProposal - Promise is rejected: Error: 2 UNKNOWN: chaincode error (status: 500, message: {"error":Could not find specified account}). When this is happening? Is it trying to do operations on those accounts before their creation? Are you creating all accounts first and then performing operations on them?

haojun:
@Ram-srini 
1. The error happens when operating on an account that has not been created. 
2. I have not change the logic of workload generation of smallbankOperations.js. So account should be created first.

I've not encountered so many errors in my test (usually less than 20 errors for 4000 txns). Can you try to lower the 'tps' of the rate controller to see if the situation gets better?

I guess it is because the process of create-account txns have not been finished while following operations were submitted. One way to prevent it is to design a two phase test which includes two Caliper test rounds, the first round is to only test creating accounts, and the second is to test other operations. Since the second test round only starts after the first round has been finished, such errors should be avoided.

Ram-srini:
@haojun small bank payload generation logic will first generate the account creation and later the other operations to those accounts. You shouldn't get the account shouldn't be exist error later. Not sure how fabric handles it. In sawtooth i am not getting any such error. In sawtooth if I submit transaction at higher rate, validator will throw exception(http status 429) of queue full error and those transactions are moving to failed state. Exception has to be handled and resubmit those transactions again.

benchmark/smallbank/smallbankOperations.js
```diff
@@ -68,7 +68,7 @@ function generateWorkload() {
     if(workload.length === txnPerBatch) {
         return workload;
     }
-    else {
```
Ram-srini: Why this changes? create account operation may be part of batch, but it should be first in the order.

haojun: Since Fabric does not support batch, txns in the same batch are submitted nearly simultaneously . There is a high chance of error if an transaction such as send_payment is wrapped in a batch with a create_account transaction and those two txns operates on the same account. The change is to avoid this.

haojun:
@Ram-srini 
Let's assume the first create_account transaction is submitted at 0s and committed on the ledger at 1s. And a send_payment transaction is submitted at 0.5s, endorsor gets the txns and try to lookup account from the ledger but it can not get it, so it will return an error to the client. Until now, fabric adapter won't do any re-submit for any error.

Does sawtooth adapter support re-submitting?

benchmark/smallbank/smallbankOperations.js
Ram-srini: Is it happening always? or with higher tps rate? Intention of small bank benchmark use case is simulate the real bank use case. if order of transactions is fixed then, it should create account first and later it perform operations on those account. For amalgamate we can do perform separate testing. Other than this we shouldn't change the behavior small bank work load generation.

haojun: It happens always. In real world, if you are using fabric, you should first create account, **waiting for the result** and then send other operations. You should not send the later operation if the creation has not been finished.

Ram-srini: Then it is limitation in Fabric, even if we are sending the transactions in the right order where account creation is first. In this case should we our benchmark test to fit to the limitation of hyperledger framework?

benchmark/smallbank/README.md
```diff
@@ -0,0 +1,9 @@
+The smallbank benchmark contains a series of transactions to simulate bank operations, including one transaction which creates accounts and others which perform banking transactions such as querying balances, depositing in accounts, payiing money, etc. The benchmark intends to provide cross-platform workload for blockchain performance testing in the bank scene.
```
feihujiang: payiing typo

src/contract/fabric/smallbank/smallbank.go
```diff
@@ -0,0 +1,263 @@
...
+	err2 = saveAccount(stub, destAccount)
+	if err1 != nil || err2 != nil {
+		return errormsg(ERROR_PUT_STATE)
+	}
+	
+	return shim.Success(nil)
+}
+
+func (t *SmallbankChaincode) Amalgamate(stub shim.ChaincodeStubInterface, args []string) pb.Response {
+	if len(args) != 3 {	// should be [amount,dest_customer_id,source_customer_id]
```
feihujiang: if len(args) != 2 {	// should be [dest_customer_id,source_customer_id]

haojun:
Summarize the current discussion:
Smallbank is a benchmark designed with *batch* which is supported by some blockchain systems such as Sawtooth. While Fabric does not support *batch*, here are some options:
1. Submit txns in batch as usual without any special handling, which may cause many txn failures as explained in the README
2. Simulate the batch behavior, a txn should not be submitted until the previous txn in the same batch has been committed (the client has gotten the notification from the event hub). That could resolve txn conflict in the same batch, but has no effect on txns across batch.
3. Modify the test flow to lower the chance of txns conflict as mentioned before, which makes the benchmark more friendly to blockchain systems without batch support.
4. Only apply this benchmark for blockchain systems which support batch. For other systems such as fabric, we should notice users that batch is not supported and recommend them to disable batch feature by setting *txPerBatch* to 1.
5. Other options??

src/contract/fabric/smallbank/smallbank.go
```diff
@@ -0,0 +1,263 @@
...
+		return errormsg(ERROR_WRONG_ARGS + " amalgamate")
+	}
+	destAccount, err1 := loadAccount(stub, args[1])
+	sourceAccount, err2 := loadAccount(stub, args[2])
+	if err1 != nil || err2 != nil {
+		return errormsg(ERR_NOT_FOUND)
+	}
+
+	// amount, _ := strconv.Atoi(args[0])
+	destAccount.SavingsBalance += sourceAccount.SavingsBalance
```
feihujiang: destAccount.CheckingBalance += sourceAccount.SavingsBalance

benchmark/smallbank/smallbankOperations.js
haojun: I've summarized the current discussion below, and will post it to contributor channel of rocketchat to see if other guys have  any suggestion.

zookeeper.out
```diff
@@ -0,0 +1 @@
+nohup: failed to run command 'java': No such file or directory
```
feihujiang: Do we need this file?

~/zookeeper/data/zookeeper_server.pid
```diff
@@ -0,0 +1 @@
+43698
```
feihujiang: Do we need this file?

benchmark/smallbank/config.json
```diff
@@ -0,0 +1,38 @@
...
+    "description" : "This is smallbank benchmark for caliper",
+    "clients": {
+      "type": "local",
+      "number": 5
+    },
+    "rounds": [{
+        "label" : "smallOperations",
+        "txNumber" : [2000],
+        "rateControl" : [{"type": "fixed-rate", "opts": {"tps" : 200}}],
+        "arguments": {"accounts":300, "txnPerBatch":10},
```
Ram-srini: If we set txnPerBatch = 1 for fabric, will it fix 'Invalid transaction: MVCC_READ_CONFLICT' error in fabric?

haojun: Unfortunately no...... But at least there will be some time interval between two adjacent submittings, so it may be better at low submitting rate.

benchmark/smallbank/smallbankOperations.js
```diff
@@ -14,7 +14,7 @@ let no_accounts=0;
 let account_array=[];
 let accounts, txnPerBatch;
 const initial_balance = 1000000;
-const operation_type = ['transact_savings','deposit_checking','send_payment','write_check','amalgamate'];
+const operation_type = ['transact_savings','deposit_checking','send_payment','write_check']; //,'amalgamate'];
```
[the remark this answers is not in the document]
haojun: OK, got it

haojun:
A new PR is created to replace this one. Closed.

PR #76, opened 2018-05-22
Fix amalgamate payload

feihujiang:


feihujiang:
Smallbank amalgamate operation doesn't need amount payload, so remove it. @Ram-srini please review.

Ram-srini:
Yes amount is not required for amalgamate.

PR #83, opened 2018-05-30
support smallbank benchmark with fabric

haojun:
Main updates:

1. Add chaincode of smallbank
2. Update fabric adaptor to support submit multiple txns using invokesmartcontract()
3. Minor changes for smallbankOperations.js, using round-robin instead of random algorithm for account selection in workload generation to avoid conflicting txns 
4. Add fabric related configuration files
5. Add a Readme doc in the smallbank folder


haojun:
@feihujiang @Ram-srini 
This is the replacement PR of the old one. I've updated codes according to previous reviews, and change the workload generation a little to avoid conflict txns as much as possible.
Please check it.

haojun:
@Ram-srini Please remove the old smallbank cc images (whose names as dev-peerxxorgx.example.com-smallbank-v0-xxx) and re-run the test

Ram-srini:
@haojun Where are those files located? I have removed files from ./network/fabric/simplenetwork/crypto-config, still i am getting same error.

Ram-srini:
@haojun Removing old docker images fixed the amalgamate error. But still there exists the problem in fabric if we generate the frequent and multiple updates on same account. Fabric rejects many transaction if there is frequent and multiple update to same block. Assume the case if tps>accounts. For example number of accounts is 30 and input tps rate 50.

haojun:
@Ram-srini 
Yes, but that's the nature of fabric, we can't avoid it 100%.

PR #85, opened 2018-06-04
keep fabric images version in configuration same with docker-compose.yaml version

JulienGuo:
keep fabric images version in configuration same with docker-compose.yaml version


haojun:
Hi @JulienGuo, you must use -s to sign off your proposal when submitting a PR

PR #94, opened 2018-06-14
add transfer to simple benchmark and fix transfer issue in chaincode

ski2per:
add transfer to simple benchmark and fix transfer issue in chaincode

### DIFF
--- a/benchmark/simple/config.json
+++ b/benchmark/simple/config.json
@@ -1,5 +1,4 @@
-{
-  "blockchain": {
+{ "blockchain": {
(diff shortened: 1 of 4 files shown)


haojun:
Sorry for the late reply. Thanks for the contribution, but first you should use 'commit -s' to sign the commitment.
And I think the simple benchmark is only an 'example' to show how to write a benchmark. We have another benchmark named smallbank which implements a more comprehensive benchmark including transferring money.

ski2per:
OK, then I will close the pull request

Issue #88, opened 2018-06-06
Two Network Directories for Fabric


ynamiki:
There are two directories for Fabric: `network/fabric` and `network/fabric-v11`.

I would like to know intentions of each directory.
`fabric` is for Fabric v1.0 only? (but I can use `fabric/simplenetwork` with Fabric v1.1)

haojun:
The original caliper as well as simplenetwork only support fabric v1.0 at the beginning, and then composer team added a new folder to include files they needed and named as fabric-v11 to distinguish with the old folder. But now caliper have been updated to support fabric v1.1, the names of those two folders may be confusing.

ynamiki:
Thanks, I understand the background.

PR #99, opened 2018-06-17
Merge network/fabric-v11 and network/fabric directories

ynamiki:
As mentioned in #88, now we don't to need separate the directories.

This patch moves contents of `network/fabric-v11` directory to `network/fabric` and update file paths in configuration files.


haojun:
@nklincoln Please confirm if composer still needs a separate folder and help to check if all corresponding paths are updated. Thanks.

nklincoln:
@haojun - the changes will not break Composer performance tests 👍 All required paths have been updated

PR #103, opened 2018-06-20
Fix rejected Fabric proposal handling

aklenik:
The error handling part of a rejected Fabric Tx proposal referenced an undefined attribute of the response.
In case of rejected proposal, an explicit `Error` object is returned as result (see [this SDK code](https://github.com/hyperledger/fabric-sdk-node/blob/bc7d7db22e209f3c2477a0023ba0e8a0201cebe6/fabric-client/lib/client-utils.js#L103)).

@haojun can you review this quick-fix?


aklenik:
The error was in my user module, so the fix works :)

PR #116, opened 2018-07-03
Node CC and metadata support, node CC examples

aklenik:
This PR addresses the following shortcomings:
* Proper support for Node.js chaincodes (CC)
* Support for CC metadata (CouchDB indices)
* Lack of Node.js CC examples

Contribution details:
* Correctly resolving CC path (relative or absolute) in case of Node CC
* Processing `metadataPath` attribute for CC install requests
* Added Node version of the `simple` benchmark (restructured the `src/contract/fabric/simple` folder a bit, changed configs accordingly)
* Added Node and Golang versions of the `marbles` CC from the `fabric-samples` repo (included a NOTE about this derivative work and a reference to the original, hopefully, this makes it license-compatible)
* Added `marbles` benchmark files for testing the corresponding CCs
* Fixed some bugs/typos in the `simple` Golang CC
* Updated chaincode-related documentation accordingly

### DIFF
--- a/benchmark/marbles/config-fabric-go.json
+++ b/benchmark/marbles/config-fabric-go.json
@@ -1,0 +1,42 @@
+{
+  "blockchain": {
+    "type": "fabric",
+    "config": "benchmark/marbles/fabric-go.json"
+  },
+  "command" : {
+    "start": "docker-compose -f network/fabric/2-org-1-peer/docker-compose-tls.yaml up -d",
+    "end" : "docker-compose -f network/fabric/2-org-1-peer/docker-compose-tls.yaml down;docker rm $(docker ps -aq);docker rmi $(docker images dev* -q)"
+  },
+  "test": {
+    "clients": {
+      "type": "local",
+      "number": 5
+    },
+    "rounds": [{
+        "label" : "init",
+        "txNumber" : [500, 500, 500],
+        "rateControl" : [{"type": "fixed-rate", "opts": {"tps" : 25}}, {"type": "fixed-rate", "opts": {"tps" : 50}}, {"type": "fixed-rate", "opts": {"tps" : 75}}],
+        "callback" : "benchmark/marbles/init.js"
+      },
+      {
+        "label" : "query",
+        "txNumber" : [15, 15],
+        "rateControl" : [{"type": "fixed-rate", "opts": {"tps" : 5}}, {"type": "fixed-rate", "opts": {"tps" : 5}}],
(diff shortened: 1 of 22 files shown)


aklenik:
@haojun Sorry, got confused since there was a refactor with the `network/fabric*` directories on the master branch.
 
Amended my commit with the correct paths according to this PR. However, during merging, the following adjustments need to be made to the network paths to be compatible with the current master branch:

`benchmark/marbles/config-fabric-[go|node].json` lines 7 and 8: `network/fabric-v11` to `network/fabric`

`benchmark/marbles/fabric-[go|node].json` lines 3, 9, 10, 13, 19, 20, 30, 44 and 51: `network/fabric-v11` to `network/fabric`

Can you apply these changes during the merge?

haojun:
I've removed '-v11' postfix but still got the error. I think the reason is that the auto-generated network name is not the same as the one defines in the yaml file. I just wondering if you can pass the test?

aklenik:
Yes, the test completes successfully for me. The network name is explicitly set in the docker-compose yamls, I haven't modified the network files. Could it be a difference between docker(-compose) versions? One version removes the - characters from the containing folder, while the other doesn't?
What if the following line is set to a network name with - characters?
https://github.com/hyperledger/caliper/blob/master/network/fabric/2-org-1-peer/docker-compose-tls.yaml#L83

haojun:
That's strange, i thought the network name is inherited from the folder name (so in my testing the network name returned by 'docker network ls' is 2-org-1-peer_default, not 2org1peer_default) or the project name. I'm not very familiar with it, so maybe I miss something.

My docker-compose version is 1.21.2. What's your version？

aklenik:
My network after running the test:
```
$ docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
53eeb36eda86        2org1peer_default   bridge              local
```

Docker(-compose) version:
```
$ docker --version
Docker version 18.03.1-ce, build 9ee9f40
$ docker-compose --version
docker-compose version 1.16.1, build 6d1ac21
```

So we get different default names with different compose versions... That's not desirable. So for now, try setting the network name explicitly to `2org1peer_default` with the `-p 2org1peer_default` compose setting. 

But I think we will need a separate PR that explicitly defines the network name in the yaml files: https://docs.docker.com/compose/networking/#specify-custom-networks

haojun:
OK. So I can approve this PR after you fix the path problem (removing -v11 postfix).

aklenik:
Done.

PR #122, opened 2018-07-10
define common TxStatus Class

haojun:
1. Define a common TxStatus class and modify all adapters accordingly.
2. Reduce memory usage of local-client.js

### DIFF
--- a/benchmark/composer/composer-micro/query-asset.js
+++ b/benchmark/composer/composer-micro/query-asset.js
@@ -31,6 +31,7 @@
 'use strict';
 
 const Util = require('../../../src/comm/util');
+const TxStatus = require('../../../src/comm/transaction');
 
 module.exports.info  = 'Query Asset Performance Test';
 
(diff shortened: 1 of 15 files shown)


haojun:
Hi @nklincoln @stinger112 @Ram-srini , please help to review this PR to check if your adapters (composer, iroha,  sawtooth) have been modified correctly and can run with no error.  

Thanks :grinning:

nklincoln:
Hi @haojun,
Whilst testing this I noticed that the change made in local-client.js has broken the ability to run composer tests.

The current while loop within `runFixedNumber` isn't behaving quite as expected. I made a change to a for loop, conditioned on the required number of invokes. I think the unexpected operation is because the global variable `txNum` relies upon an update performed once the `cb.run()` has completed, which means it is possible to send in excess of the number of intended invokes if the cb.run() duration is in excess of the apply rate control duration ... as is current using a while loop. My change is:

```
for(let i = 0 ; i < msg.numb ; i++) {
        promises.push(cb.run().then((result) => {
            addResult(result);
            return Promise.resolve();
        }));
        await rateControl.applyRateControl(start, txNum, results);
    }
```

And the test completed in combination with your changes in this PR

aklenik:
Although the Fabric query API is hardcoded right now, in the `src/fabric/e2eUtils.js` file before L771 a `txStatus.SetVerification(true)`, and before L781 a `txStatus.SetVerification(false)` is missing (although the latter is the default behavior). But it's not critical since this part needs to be implemented properly anyway.

haojun:
@nklincoln, 

By #66 a 'submitCallback' is injected into the output of getcontext() to count the actual tx submitting(see from https://github.com/hyperledger/caliper/blob/6131f4fca8881d9bb2f3a0fdc829b7df7036659e/src/comm/client/local-client.js#L165).

Sorry I forgot to change composer accordingly, and I think that's the real cause of the composer test failure.

Can you check and help to fix it? Thanks.

nklincoln:
@haojun, ok - thanks... in which case the 'quick fix' there is to modify `composer.js` `submitTransaction()` method to include:

```
submitTransaction(connection, transaction) {
        let invoke_status = new TxStatus(transaction.getIdentifier());
        if(connection.engine) {
            connection.engine.submitCallback(1);
        }
       return connection.submitTransaction(transaction)
...
```

This is likely to have emerged due to the bad naming convention used in the composer plug, which does not retain the `invokeSmartContract` convention used by the other plugins ... happy to raise a PR and fix this if desired.

haojun:
@nklincoln done

haojun:
Checked with Sawtooth and Iroha

JulienGuo:
thanks @nklincoln 
Can you help me to add a new feature of pidrate, I want change the invoke timeout from a constant value to a variable value , and Please make the value related to sleep time.

PR #124, opened 2018-07-12
Split caliper into transaction sender and event listener

nitesh7sid:
PR for spitting caliper process into transaction sender and event listener.
## Short Description
Current version of the caliper relies on a single threaded process to send the transactions as well as listen for the transaction confirmation event. This can cause an issue for clients with higher send rates since node.js is itself a single threaded. The callback of the eventhub may get queued in the event loop and hence may result in incorrect confirmation time.

To tackle this issue, caliper can be split into 2 sub processes namely,
* sender who is just responsible for generating the transaction load
* listener which connects to the event peer, listens for the block event, note the timestamp and then publish it to Apache Kafka MQ. 

A persistence layer is required so that after the sender has finished generating the load, it can then consume blocks and associated confirmation time from the Queue and compute `time_final` property for every transaction.

### DIFF
--- a/src/comm/bench-flow.js
+++ b/src/comm/bench-flow.js
@@ -289,8 +289,6 @@ function execAsync(command) {
  * @param {String} networkFile path of the blockchain configuration file
  */
 module.exports.run = async function(configFile, networkFile) {
-    let localConfig = require(configFile);
-    configurationType = localConfig.test.clients.WITH_MQ;
     logger.info('####### Caliper Test ######');
     absConfigFile  = Util.resolvePath(configFile);
     absNetworkFile = Util.resolvePath(networkFile);
@@ -301,6 +299,7 @@ module.exports.run = async function(configFile, networkFile) {
     demo.init();
 
     let configObject = require(absConfigFile);
+    configurationType = configObject.test.clients.WITH_MQ;
 
     try {
         if (configObject.hasOwnProperty('command') && configObject.command.hasOwnProperty('start')) {


haojun:
Hi @nitesh7sid ,
Thanks for this contribution, I'll try it next week. 

Before that, I have some quick questions and comments:
1. I'd like to make it an 'optional' feature so people who don't want to install mq can still use the old method. And it also can help people to compare results of using different ways. An addiontional parameter can be added in the configuration file to specify this.
2. It would be better to start the listener in bench-flow.js instead of main.js
3. Is there only one global listener for multiple local or remote (zookeeper mode) senders (clients)? Would the listener itself be a bottleneck?

haojun:
4. I just noticed the confirmation results are fetched at the end of the round, which means it's impossible to calculate the realtime metrics as we do now by the txUpdateInter object. Right?

nitesh7sid:
Hi @haojun 

- yes, there will be only one global listener for multiple local or remote (zookeeper mode) senders. I don't think listener will be a bottleneck as it will be running entirely on a separate process. I have also tested with multiple clients at higher send rates and didn't encounter any bottleneck issues yet.

- yes, confirmation results are fetched at the end of the round. so it will not be possible to calculate the realtime metrics.

haojun:
@nitesh7sid 
Since all results are maintained by the listener, there it no need for clients to fetch results from the listener then send them to the master process (bench-flow.js). The listener itself can calculate the statistics and send them to the master process. And the listener can also calculate the realtime metrics.

nitesh7sid:
@haojun listener process only notes the confirmation time of every block arrived and then publishes to kafka. After the round has finished, the local-client.js makes a call to the function "getTransactionConfirmationTime" to consume the blocks and confirmation time for every txn. https://github.com/nitesh7sid/caliper/blob/master/src/comm/client/local-client.js#L187
listener process doesn't have the resultsArray which includes time-created, time_endorse, etc.

nitesh7sid:
@haojun how about this:
a) create a third process which runs in parallel to fetch the confirmation time from MQ, and also read the pending transactions from the "sender" and emit the statistics?

haojun:
@nitesh7sid 

It sounds good. I'd like to review after you finished the modification aforementioned (add the feature as optional and support calculating realtime metrics)

nitesh7sid:
Hi @haojun  I have made the changes as suggested above.

haojun:
@nitesh7sid , thanks. Please resolve the conflicting so that I can merge and test the code.

nitesh7sid:
@haojun what is the procedure to run the linting?can u please provide the steps ?thanks.

haojun:
@nitesh7sid , run `npm run lint` at the root folder.

aklenik:
@nitesh7sid @haojun If you allow me some remarks before merging:
1) The `package.json` file shouldn't contain dependencies that are specific to one of the blockchain platforms (in this case: `fabric-ca-client`, `fabric-client`, `grpc`, `kafka-node` )
2) The source files should be under the `src/` hierarchy. Since this extension is not Fabric specific, I suggest `src/listener/`
3) The kafka compose file should be in `network/kafka-listener`, for example. 
4) Configurations shouldn't be repeated in multiple files, for example, the `peerEventUrl` is also available in the network config files (same for the other attributes)
5) Are we sure that listening to one peer for events is enough? What if that peer crashes? The transaction still can be successfully committed on other peers. Plus if we want to measure how fast transactions are committed across the network, we will need all event sources (see the PSWG docs about these questions)
6) I think it would be better to set `WITH_MQ` as an environment variable right now. There should be proper configuration support for Caliper soon (see issue #138). And it should also be named `CALIPER_CORE_USEMQ` :)

That's all for now. I'll deep dive into the code in more details tonight, and write an actual review if you're interested.

src/gui/src/demoOptional.js
```diff
@@ -0,0 +1,305 @@
...
+        var submitted_transactions = updates[i].committed
+    
+        for (let j =0; j < submitted_transactions.length; j++) {
+            var transactionStatus = submitted_transactions[j].status
+            var TransactionStatus = new TxStatus(transactionStatus.id, transactionStatus.status, transactionStatus.time_create, transactionStatus.time_final, 
+                transactionStatus.result, transactionStatus.verified, transactionStatus.flags, transactionStatus.error_messages);
+                
+            if (cachedEvents.get(TransactionStatus.GetID()) == undefined) {
+                cachedEvents.set(TransactionStatus.GetID(), TransactionStatus)
+                if (label == 'query') {
```
haojun: 1. I think it's not proper to use any test case specific informations in a general benchmark module to do some patches. The tester can use any label name and the semantic is unknowable to a general benchmark module.
2. Is it necessary to define a new demo file? Can we just use the old update mechanism?

nitesh7sid:
Hi @haojun thanks for the comments! I am not able to reproduce the same error. It worked fine for me. Did you start the kafka cluster and change the IP in the listener-config.json file ?

benchmark/simple/config.json
```diff
@@ -8,23 +8,22 @@
     "end" : "docker-compose -f network/fabric/simplenetwork/docker-compose.yaml down;docker rm $(docker ps -aq)"
   },
   "test": {
-    "name": "simple",
-    "description" : "This is an example benchmark for caliper, to test the backend DLT's performance with simple account opening & querying transactions",
+    "WITH_MQ": true,
```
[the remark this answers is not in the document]
nitesh7sid: if the flag is not set, the tool will default to non MQ mode.

package.json
```diff
@@ -23,12 +23,17 @@
     "cbor": "^3.0.3",
     "commander": "^2.11.0",
     "dockerode": "^2.5.0",
+    "fabric-ca-client": "^1.1.0",
```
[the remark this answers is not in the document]
nitesh7sid: agree. I will make sure they are removed from the package.json file

haojun:
@nitesh7sid , the kafak cluster has been started (kafaka0/1/2 and zookeeper0/1/2 can be found by 'docker ps') and I also changed the listener-config.json as below:
```
{
    "broker_urls": "localhost:9092,localhost:7092,localhost:8092",
"zk_url": "localhost:2181",
"topic": "client-b",
"peerOrg":"org1",
"peerEventUrl":"grpcs://localhost:7053",
"peerEventHostnameOverride":"peer0.org1.example.com",
"peerEventTlscaPath":"network/fabric/simplenetwork/crypto-config/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt"
 }
``` 

Now the error becomes ```Error while publishing block in kafka [ 'LeaderNotAvailable' ]```.

nitesh7sid:
oh, it looks like you are running the kafka cluster on the same machine where fabric and caliper is running. I run the kafka cluster on a separate vm or machine. since kafka cluster is composed of 6 docker containers, its better to allocate a separate resource for it.

haojun:
@nitesh7sid , I started the kafka on another machine(139.159.243.213) as you suggested, now the error disappeared, but the test process is blocked at:
```
# ##### testing 'open' ######
----test round 0----open
started monitor successfully
Info: client 77688 start test runFixedNumber():opening accounts
***** { type: 'fixed-rate', opts: { tps: 50 } }
Info: client 77689 start test runFixedNumber():opening accounts
Info: client 77697 start test runFixedNumber():opening accounts
***** { type: 'fixed-rate', opts: { tps: 50 } }
***** { type: 'fixed-rate', opts: { tps: 50 } }
Info: client 77683 start test runFixedNumber():opening accounts
***** { type: 'fixed-rate', opts: { tps: 50 } }
```

Here is the new listener-config.json:
```
{
    "broker_urls": "139.159.243.213:9092,139.159.243.213:7092,139.159.243.213:8092",
     "zk_url": "139.159.243.213:2181",
     "topic": "client-b",
     "peerOrg":"org1",
     "peerEventUrl":"grpcs://localhost:7053",
     "peerEventHostnameOverride":"peer0.org1.example.com",
     "peerEventTlscaPath":"network/fabric/simplenetwork/crypto-config/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt"
 } 
```
PS: which ip address should I use for KAFKA_ADVERTISED_HOST_NAME, since all kafka nodes are on the same machine, i guess localhost is ok.

haojun:
@nitesh7sid It works now, thanks. I'll do more test later on zk mode and other blockchains.

2 comments after the quick test:

1. The console log of the realtime metrics refreshes too quickly. 
For reference, demo.js will generate an output every one second(base on the value of the global demoInterval), and the output will also be saved for a GUI to draw performance charts whose x-axis scale is also one second. 
I strongly recomend to modify demo.js to support mq mode instead of introducing a new demoOptional.js. There are many tricks on the demo, and two implementation would make later modification more difficult :(

2. The process didn't quit after the test, i had to use ctrl-c to kill it.

nitesh7sid:
Thanks @aklenik for your time in reviewing the PR.

@aklenik
1)	Agreed. Removing the fabric-client, fabric-ca-client, grpc and kafka-node dependencies from package.json
2)	Agreed. Moving the listener module src/listener directory and kafka compose set up file at network/ directory
3)	The peerEventUrl configurations are repeated in listener-config.json file to give users the flexibility to choose different peers to listen to events for overall load balancing. We recommend each listener connect to a single peer to avoid receiving duplicate tx confirmation events. This problem adds extra  processing burden on the client and worsens when additional channels are present. Assuming that the tool will be run in controlled settings for benchmarking, this is the design choice we made to keep the event listener light-weight. 

@haojun 
1)	In demoOptional.js file there is a check made to determine the type of benchmark i.e. invoke or query. If it is query, then there would be no blocks in the MQ and hence demoOptional.js would not have to connect to the kafka MQ. To remove the use of  test case specific information, the best approach will be to accept a Boolean argument in the startWatch function and based on the argument determine whether to connect to MQ or not. For this to happen, the `WITH_MQ` flag should be set for every test case in the rounds array in config.json file. So before starting any round, there can be a check made in the bench-flow.js - if the `WITH_MQ` is true or false and depending on that it can instruct the startWatch function whether to connect to MQ or not.
2)	With this approach we can eliminate the global `WITH_MQ` flag. While starting the benchmark, we can accept an ENV variable for Caliper tool to know that it has to now spawn a listener process.
3)	I will update the readme as suggested.
4)	Currently kafka mode is supported only for Fabric only. Zk mode is still an experimental feature and I couldn’t find sufficient documentation on how to run ZK mode to test the newly added functionality. We will need to work together to make it work for the Zk mode. 
5)	demoOptional.js prints the metrics information once per second only. That code has not changed. It looks like the `success` counter gets incremented very fast, because with the new design, the timestamps have a higher accuracy as the listener is only listening to events and timestamping. That is the intended improvement that this design is expected to provide.

haojun:
@nitesh7sid thanks for you reply, below is my thoughts:

1. I'm not sure moving WITH_MQ flag to each test round is a good idea. The tester may not be a skilled person who understands the difference between invoking and querying. And how to set the flag if a test round do both invoking and querying (please note people can do whatever they want in the callback script)? It's very easy to distinguish invoking and querying, they are using different functions, one is invokesmartcontract() and another is querystate(). So my idea to solve the problem is:
(1). Do not block results from callback.run(). You can change the invokesmartcontract() to return an empty array if mq is enabled.
(2). Reuse the current txUpdated and testResult message for the block handler (the one which parses the block and generates the statistics, i don't know how you call it) and save the results in the buffer in client.js as usual. 
(3). It may be better to fork the listener from client.js, for example in the client.init() 
So the client.js can get the statistics from local clients in case of querying as the old way and block handler in case of invoking with mq. And you don't even need to modify the demo.js or processing result in bench-flow.js, since the buffer in client.js is used as usual.

4. Got it. Then you should add some words in the readme to explain it only support local test with fabric now. And since it only support local client, it may be better to move the flag into 'clients' property with type=local

5. From what I saw, the console prints dozens of statistics lines every second.

6. IMHO, copying & modifying an old file to produce an optional version (such as demoOptional.js and e2eUtils_withMQ.js) is not a good way. Most functions of those files are the same as old ones. From a maintainer's point of view, it is bad for project maintance.

nitesh7sid:
Hi @haojun - I am trying to move the code from demoOptional.js to client-util.js file. I could achieve the following:
1) add a new status field 'NeedVerifyWithMQ' in the TxStatus object and set it true if MQ mode is running.
2) from local-client.js send the un-verified transactions to the client-util.js using a different type of message.
3) In client-util.js, handle the new message and push the un-verified transactions in some buffer.
4) In the init function of client.js, invoke a function from client-util.js 'cosumeEvents' which connects to kafka and stores the confirmation time in a hash map. the hash map is defined in client-util.js
4) in the startTest function of client-util.js, start the interval object to traverse the unconfirmed transactions buffer and find the time_final for each transactions. But now i am facing couple of issues:

* before all the transactions are verified from the buffer, the round exists and the tool starts with the next round. how should i handle this?
* how do I confirm that the round is over and clear the interval object ?

below is the code of client-util.js. I am not sure if I am missing anything. Could you please check ? Thanks
```/**
* Copyright 2017 HUAWEI. All Rights Reserved.
*
* SPDX-License-Identifier: Apache-2.0
*
*/


'use strict';
const log          = require('../util.js').log;
let processes  = {}; // {pid:{obj, promise}}
let txUpdateTime = 1000;
const kafka = require('kafka-node');
const listener_config = require("../../listener/listener-config.json");
let confirmedTransactions = [];
let cachedEvents = new Map();
let unConfirmedTransactions;
var txUpdateInter = null;
var globalConsumer;
const TxStatus = require('../transaction.js');
var updateTail;

function _consumeEvents(){
    var Consumer = kafka.Consumer;
    var KafkaClient = new kafka.KafkaClient({ kafkaHost: listener_config.broker_urls, requestTimeout: 300000000 });
    var options = {
    autoCommit: true,
    fetchMaxWaitMs: 1000,
    fetchMaxBytes: 5120 * 5120,
    encoding: 'buffer',
    groupId: "groupID" + Math.floor(Math.random() * Math.floor(100))};
 
    var topics = [{
        topic: listener_config.topic
    }];
 
    var consumer = new Consumer(KafkaClient, topics, options);
    globalConsumer = consumer;
    consumer.on('message', function (message) {
        var buf = new Buffer(message.value); // Read string into a buffer.
        var data = buf.toString('utf-8')
        var block = JSON.parse(data).block
        
        for (var index = 0; index < block.data.data.length; index++) {
            var channel_header = block.data.data[index].payload.header.channel_header;
            var transaction_id = channel_header.tx_id
            var confirmation_time = JSON.parse(data).validTime;
            if (cachedEvents.get(transaction_id)  == undefined)
           {
                cachedEvents.set(transaction_id, confirmation_time)
           }
            else if (cachedEvents.get(transaction_id) != undefined){
 
                var transactionObject = cachedEvents.get(transaction_id);
                transactionObject.Set('time_final', confirmation_time);
                transactionObject.SetVerification(true);
                transactionObject.Set('status','success');
                cachedEvents.set(transaction_id, transactionObject);
                confirmedTransactions.push(transactionObject);
            }
        }
    });
 
    consumer.on('error', function(err){   
        // retry if unable to connect to kafka
        _consumeEvents()
    })
} 
module.exports._consumeEvents = _consumeEvents;
/**
 * Call the Promise function for a process
 * @param {String} pid pid of the process
 * @param {Boolean} isResolve indicates resolve(true) or reject(false)
 * @param {Object} msg input for the Promise function
 */
function setPromise(pid, isResolve, msg) {
    let p = processes[pid];
    if(p && p.promise && typeof p.promise !== 'undefined') {
        if(isResolve) {
               
            p.promise.resolve(msg);
        }
        else {
            p.promise.reject(msg);
        }
    }
}

/**
 * Push test result from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data test result
 */
function pushResult(pid, data) {
    let p = processes[pid];
    if(p && p.results && typeof p.results !== 'undefined') {
        p.results.push(data);
    }
}

/**
 * Push update value from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data update value
 */
function pushUpdate(pid, data) {
    let p = processes[pid];
    if(p && p.updates && typeof p.updates !== 'undefined') {

        p.updates.push(data);
    }
}

/**
 * Push update value from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data update value
 */
function pushUpdateForMQ(pid, data) {
    unConfirmedTransactions.push(data);
    /*let p = processes[pid];
    if(p && p.updatesWithMQ && typeof p.updatesWithMQ !== 'undefined') {
        p.updatesWithMQ = data;
    }*/
}


/**
 * Launch a child process to do the test
 * @param {Array} updates array to save txUpdate results
 * @param {Array} results array to save the test results
 */
function launchClient(updates, results) {
    let path = require('path');
    let childProcess = require('child_process');
    let child = childProcess.fork(path.join(__dirname, 'local-client.js'));
    let pid   = child.pid.toString();
    processes[pid] = {obj: child, results: results, updates: updates};

    child.on('message', function(msg) {
        if(msg.type === 'testResult') {

            pushResult(pid, msg.data);
            setPromise(pid, true, msg.data);
        }
        else if(msg.type === 'error') {
            setPromise(pid, false, new Error('Client encountered error:' + msg.data));
        }
        else if(msg.type === 'txUpdated') {
            pushUpdate(pid, msg.data);
        }
        else if(msg.type === 'txUpdatedWithMQ') {
            pushUpdateForMQ(pid, msg.data);
        }
    });

    child.on('error', function(){
        setPromise(pid, false, new Error('Client encountered unexpected error'));
    });

    child.on('exit', function(){
        log('Client exited');
        setPromise(pid, true, null);
    });
}

/**
 * Start a test
 * @param {Number} number test clients' count
 * @param {JSON} message start message
 * @param {Array} clientArgs each element contains specific arguments for a client
 * @param {Array} updates array to save txUpdate results
 * @param {Array} results array to save the test results
 * @return {Promise} promise object
 */

function update() {
    console.log("Pending transactions length", unConfirmedTransactions.length)
    var data = [];
    var len  = unConfirmedTransactions.length;
    if(len > updateTail) {
        data = unConfirmedTransactions.slice(updateTail, len);
        updateTail = len;
    }
     demoRefreshData(data);
}

function demoRefreshData(updates) {

    for (let i = 0; i < updates.length; i++) {
        var sub = updates[i].submitted;
        var submitted_transactions = updates[i].committed;
        for (let j =0; j < submitted_transactions.length; j++) {
            var transactionStatus = submitted_transactions[j].status
            var TransactionStatus = new TxStatus(transactionStatus.id, transactionStatus.status, transactionStatus.time_create, transactionStatus.time_final, 
                transactionStatus.result, transactionStatus.verified, transactionStatus.flags, transactionStatus.error_messages);
                
            if (cachedEvents.get(TransactionStatus.GetID()) == undefined) {
                // make an entry into map only if it is not a query OR the error flag == 0
                if (TransactionStatus.GetFlag() == 0) {
                    cachedEvents.set(TransactionStatus.GetID(), TransactionStatus)
                } else {
                    confirmedTransactions.push(TransactionStatus)
                }
            }
            else {
                TransactionStatus.Set('time_final', cachedEvents.get(TransactionStatus.GetID()))
                TransactionStatus.SetVerification(true)
                TransactionStatus.Set('status', 'success')
                cachedEvents.set(TransactionStatus.GetID(), TransactionStatus)
                confirmedTransactions.push(TransactionStatus)
                //getDefaultStats(totalSubmitted, confirmedTransactions, label)
            }   
        }  
    }
    console.log("confirmed", confirmedTransactions.length)
    //newStats = blockchain.getDefaultTxStats(confirmedTransactions, false);
   //getDefaultStats(totalSubmitted, label)
}

function startTest(number, message, clientArgs, updates, results) {
    let count = 0;
    for(let i in processes) {
        i;  // avoid eslint error
        count++;
    }
    if(count === number) {
        txUpdateInter = null;  
        unConfirmedTransactions = [];
        updateTail = 0;
    if(txUpdateInter === null) {
        // start a interval to query updates
        txUpdateInter = setInterval(update, 1 * 1000);
    } 

        // already launched clients
        let txPerClient;
        if (message.numb) {
            // Run specified number of transactions
            txPerClient  = Math.floor(message.numb / number);

            // trim should be based on client number if specified with txNumber
            if (message.trim) {
                message.trim = Math.floor(message.trim / number);
            }

            if(txPerClient < 1) {
                txPerClient = 1;
            }
            message.numb = txPerClient;
        } else if (message.txDuration) {
            // Run for time specified txDuration based on clients
            // Do nothing, we run for the time specified within message.txDuration
        } else {
            return Promise.reject(new Error('Unconditioned transaction rate driving mode'));
        }

        message.clients = number;

        let promises = [];
        let idx = 0;

        for(let id in processes) {
            let client = processes[id];
            let p = new Promise((resolve, reject) => {
                client.promise = {
                    resolve: resolve,
                    reject:  reject
                };
            });
            promises.push(p);
            client.results = results;
            client.updates = updates;
            message.clientargs = clientArgs[idx];
            message.clientIdx = idx;
            idx++;
            client.obj.send(message);
        }

        return Promise.all(promises).then(()=>{
            // clear promises
            
                for(let client in processes) {
                    delete client.promise;
                }   
                return Promise.resolve();
         
        }).catch((err)=>{
            return Promise.reject(err);
        });

    }

    // launch clients
    processes = {};
    for(let i = 0 ; i < number ; i++) {
        launchClient(updates, results);
    }

    // start test
    return startTest(number, message, clientArgs, updates, results);
}
module.exports.startTest = startTest;

/**
 * Send message to all child processes
 * @param {JSON} message message
 * @return {Number} number of child processes
 */
function sendMessage(message) {
    for(let pid in processes) {
        processes[pid].obj.send(message);
    }
    return processes.length;
}
module.exports.sendMessage = sendMessage;

/**
 * Stop all test clients(child processes)
 */
function stop() {
    for(let pid in processes) {
        processes[pid].obj.kill();
    }
    processes = {};
}
module.exports.stop = stop;
```

haojun:
@nitesh7sid , you can create a promise object to indicate whether the all the transactions are verified from the buffer, for example (I do it in _startLocalTest(), you can also do it in startest())
_startLocalTest(message, clientArgs) {
        let testfinished = false;                   // indicator to show whether the clients have finished the test
        let mqPromise = new Promise(...);  // create a new promise
        ......                                                // start an interval to check the buffer, if testfinished is true and there is no pending txns, resolve mqPromise and kill itself
        message.totalClients = this.number;
        return clientUtil.startTest(this.number, message, clientArgs, this.updates.data, this.results)
                      .then(()=>{
                           testfinished = true;
                           return mqPromise;
                      })
                      .then(()=>{
                           ... // push statistcs of txns confirmed by mq into this.results array
                      })
                      .catch((err)=>{
                           ... // exception, kill the interval obj, etc.
                      });
}

Some questions about client_util.js
1.  It seems that the txns in unConfirmedTransactions are not removed even after they have already been verified in demoRefreshData(), it may cause a txn to be checked multiple times
2. The statistics of new confirmed txns should be pushed to updates array
3. Names of some functions such as demoRefreshData are not suitable.

nitesh7sid:
Thanks @haojun . Let me give a shot. Also, the thing's you mentioned will b taken care of

nitesh7sid:
Hi @haojun I was able to get things working but found one issue in the txUpdate function of local-client.js and for which I would need your suggestions:
```
function txUpdate() {
    let newNum = txNum - txLastNum;
    txLastNum += newNum;

    let newResults = results.slice(0);
    results = [];
    let bufferToSend = [];
   
    if(newResults.length === 0 && newNum === 0) {
        return;
    }

    let newStats;
    if(newResults.length === 0) {
        newStats = bc.createNullDefaultTxStats();
    }
    else {
        newStats = blockchain.getDefaultTxStats(newResults, false);
        for (var i =0; i < newResults.length; i++){
            var txObject = newResults[i];
            // it is running in kafka mode or it is a query
            if (txObject.GetneedVerifyWithMQFlag()){
                bufferToSend.push(txObject)
            }
        }
    }

    if (bufferToSend.length !=0) {
        process.send({type: 'txUpdatedWithMQ', data: {submitted: newNum, committed: bufferToSend}}); 
    }
    else if (bufferToSend.length == 0) {
       
        process.send({type: 'txUpdated', data: {submitted: newNum, committed: newStats}});

        if (resultStats.length === 0) {
            switch (trimType) {
            case 0: // no trim
                resultStats[0] = newStats;
                break;
            case 1: // based on duration
                if (trim < (Date.now() - startTime)/1000) {
                    resultStats[0] = newStats;
                }
                break;
            case 2: // based on number
                if (trim < newResults.length) {
                    newResults = newResults.slice(trim);
                    newStats = blockchain.getDefaultTxStats(newResults, false);
                    resultStats[0] = newStats;
                    trim = 0;
                } else {
                    trim -= newResults.length;
                }
                break;
            }
        } else {
            resultStats[1] = newStats;
            bc.mergeDefaultTxStats(resultStats);
        }
    }
}
```
The issue basically is =>while running in MQ mode, sometimes the count of submitted transactions received in client-utils.js exceeds the actual transactions being sent. This is happening because both the "if" and "else" block in the txUpdate function of local-client.js is executed. So basically in txUpadte function I am traversing the newResults array and determining which txns are to supposed to be processed with MQ and accordingly populating the "bufferToSend" array. If bufferToSend is empty then send message to client-util.js using `txUpdated` as usual and if its not empty then send a new message 'txUpdatedWithMQ' to the client-util.js. So when the caliper is running in non-mq mode, it might so happen that the **If** block as well as **else** block is executed and hence the client-util.js will get messages from txUpdated as well as txUpdatedWithMQ. this results in client-util.js getting extra count of submitted transactions. one from txUpdatedWithMQ and another from txUpdated.  Hope I am clear in conveying the issue. let me know if you need more clarity on that.
Below is the code of client-util.js
```


'use strict';
const log          = require('../util.js').log;
let processes  = {}; // {pid:{obj, promise}}
let txUpdateTime = 1000;
const kafka = require('kafka-node');
const listener_config = require("../../listener/listener-config.json");
let confirmedTransactions = [];
let cachedEvents = new Map();
let unConfirmedTransactions = [];
var txUpdateInter = null;
var globalConsumer;
const TxStatus = require('../transaction.js');
var updateTail;
const bc   = require('../blockchain.js');
let path = require('path');
const blockchain = new bc(path.join(__dirname, '../../../', 'benchmark/simple/fabric'));
let testfinished = false; 
var newNum = 0;
let global_pid;
let confirmTail = 0;
var totalTransactionsRecieved = 0;
var totalTransactionsCommitted = 0;

function _consumeEvents(){

    var Consumer = kafka.Consumer;
    var KafkaClient = new kafka.KafkaClient({ kafkaHost: listener_config.broker_urls, requestTimeout: 300000000 });
    var options = {
    autoCommit: true,
    fetchMaxWaitMs: 1000,
    fetchMaxBytes: 5120 * 5120,
    encoding: 'buffer',
    groupId: "groupID" + Math.floor(Math.random() * Math.floor(100))};
 
    var topics = [{
        topic: listener_config.topic
    }];
 
    var consumer = new Consumer(KafkaClient, topics, options);
    globalConsumer = consumer;
    consumer.on('message', function (message) {
        
        var buf = new Buffer(message.value); // Read string into a buffer.
        var data = buf.toString('utf-8')
        var block = JSON.parse(data).block
        
        for (var index = 0; index < block.data.data.length; index++) {
            var channel_header = block.data.data[index].payload.header.channel_header;
            var transaction_id = channel_header.tx_id
            var confirmation_time = JSON.parse(data).validTime;
            if (cachedEvents.get(transaction_id)  == undefined)
            {
                cachedEvents.set(transaction_id, confirmation_time)
            }
           else if (cachedEvents.get(transaction_id) != undefined) {
               
                var transactionObject = cachedEvents.get(transaction_id);
                transactionObject.Set('time_final', confirmation_time);
                transactionObject.SetVerification(true);
                transactionObject.Set('status','success');
                cachedEvents.set(transaction_id, transactionObject);
                confirmedTransactions.push(transactionObject);
                totalTransactionsCommitted++;
            
            }
        }
    });
 
    consumer.on('error', function(err){   
        // retry if unable to connect to kafka
        _consumeEvents()
    });
   
} 
module.exports._consumeEvents = _consumeEvents;
/**
 * Call the Promise function for a process
 * @param {String} pid pid of the process
 * @param {Boolean} isResolve indicates resolve(true) or reject(false)
 * @param {Object} msg input for the Promise function
 */
function setPromise(pid, isResolve, msg) {
    let p = processes[pid];
    if(p && p.promise && typeof p.promise !== 'undefined') {
        if(isResolve) {            
            p.promise.resolve(msg);
        }
        else {
            p.promise.reject(msg);
        }
    }
}

/**
 * Push test result from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data test result
 */
function pushResult(pid, data) {
    let p = processes[pid];
    if(p && p.results && typeof p.results !== 'undefined') {
        p.results.push(data);
    }
}

/**
 * Push update value from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data update value
 */
function pushUpdate(pid, data) {
    let p = processes[pid];
    if(p && p.updates && typeof p.updates !== 'undefined') {

        p.updates.push(data);
    }
}

/**
 * Push update value from a child process into the global array
 * @param {String} pid pid of the child process
 * @param {Object} data update value
 */
function pushUpdateForMQ(pid, data) {
    totalTransactionsRecieved += data.submitted;
    unConfirmedTransactions.push(data); 
}


/**
 * Launch a child process to do the test
 * @param {Array} updates array to save txUpdate results
 * @param {Array} results array to save the test results
 */
function launchClient(updates, results) {
    //let path = require('path');
    let childProcess = require('child_process');
    let child = childProcess.fork(path.join(__dirname, 'local-client.js'));
    let pid   = child.pid.toString();
    processes[pid] = {obj: child, results: results, updates: updates};

    child.on('message', function(msg) {
        if(msg.type === 'testResult') {
            pushResult(pid, msg.data);
            setPromise(pid, true, msg.data);
        }
        else if(msg.type === 'error') {
            setPromise(pid, false, new Error('Client encountered error:' + msg.data));
        }
        else if(msg.type === 'txUpdated') {
            pushUpdate(pid, msg.data);
        }
        else if(msg.type === 'txUpdatedWithMQ') {
            global_pid = pid;
            pushUpdateForMQ(pid, msg.data);
        }
    });

    child.on('error', function(){
        setPromise(pid, false, new Error('Client encountered unexpected error'));
    });

    child.on('exit', function(){
        log('Client exited');
        setPromise(pid, true, null);
    });
}

/**
 * Start a test
 * @param {Number} number test clients' count
 * @param {JSON} message start message
 * @param {Array} clientArgs each element contains specific arguments for a client
 * @param {Array} updates array to save txUpdate results
 * @param {Array} results array to save the test results
 * @return {Promise} promise object
 */

function update() {
    var data = [];
    var len  = unConfirmedTransactions.length;
    if(len > updateTail) {
        data = unConfirmedTransactions.slice(updateTail, len);
        updateTail = len;
    }
      verifyUnconfirmedTransactions(data);    
}

function verifyUnconfirmedTransactions(updates) {
  
        newNum = 0;
        for (let i = 0; i < updates.length; i++) {
            var sub = updates[i].submitted;
            newNum += sub;
            var submitted_transactions = updates[i].committed;

            for (let j =0; j < submitted_transactions.length; j++) {
                var transactionStatus = submitted_transactions[j].status
                var TransactionStatus = new TxStatus(transactionStatus.id, transactionStatus.status, transactionStatus.time_create, transactionStatus.time_final, 
                    transactionStatus.result, transactionStatus.verified, transactionStatus.flags, transactionStatus.error_messages);
                
                if (cachedEvents.get(TransactionStatus.GetID()) == undefined) {
                    
                    // make an entry into map only if it is not a query OR the error flag == 0
                    if (TransactionStatus.GetFlag() == 0) { 
                        cachedEvents.set(TransactionStatus.GetID(), TransactionStatus)
                    } else {
                        confirmedTransactions.push(TransactionStatus);
                        totalTransactionsCommitted++;
                    }
                }
                else  {
                    TransactionStatus.Set('time_final', cachedEvents.get(TransactionStatus.GetID()))
                    TransactionStatus.SetVerification(true)
                    TransactionStatus.Set('status', 'success')
                    cachedEvents.set(TransactionStatus.GetID(), TransactionStatus)
                    confirmedTransactions.push(TransactionStatus);
                    totalTransactionsCommitted++;
                }   
            }
        }
            let newResults = [];
            var len  = confirmedTransactions.length;
            if(len > confirmTail) {
                newResults = confirmedTransactions.slice(confirmTail, len);
                confirmTail = len;
            }
                let newStats = blockchain.getDefaultTxStats(newResults, false);
                var dataToUpdate = {submitted: newNum, committed: newStats};
                pushUpdate(global_pid, dataToUpdate);  
               
    }
 
function updateResults(withMQ) {
    if (withMQ && unConfirmedTransactions.length != 0) {
        return new Promise(function(resolve, reject){
                (function wait(){
                    if (totalTransactionsCommitted ==  totalTransactionsRecieved && testfinished == true) { 
                        let finalStats = blockchain.getDefaultTxStats(confirmedTransactions, false);
                        pushResult(global_pid, finalStats);
                        clearInterval(txUpdateInter);
                        return resolve();
                    }
                    setTimeout(wait, 5000);
                })();
        })    
    }
    else {
        return new Promise(function(resolve, reject){
                resolve();
        }) 
    }
}

function startTest(number, message, clientArgs, updates, results, withMQ) {

    testfinished = false;   
    let count = 0;
    for(let i in processes) {
        i;  // avoid eslint error
        count++;
    }
    if(count === number) {
        
        if (withMQ) {
            txUpdateInter = setInterval(update, txUpdateTime);
            updateTail = 0;
            confirmTail = 0;
            unConfirmedTransactions = [];
            confirmedTransactions = [];
        }
       
        // already launched clients
        let txPerClient;
        if (message.numb) {
            // Run specified number of transactions
            txPerClient  = Math.floor(message.numb / number);

            // trim should be based on client number if specified with txNumber
            if (message.trim) {
                message.trim = Math.floor(message.trim / number);
            }

            if(txPerClient < 1) {
                txPerClient = 1;
            }
            message.numb = txPerClient;
        } else if (message.txDuration) {
            // Run for time specified txDuration based on clients
            // Do nothing, we run for the time specified within message.txDuration
        } else {
            return Promise.reject(new Error('Unconditioned transaction rate driving mode'));
        }

        message.clients = number;

        let promises = [];
        let idx = 0;

        for(let id in processes) {
            let client = processes[id];
            let p = new Promise((resolve, reject) => {
                client.promise = {
                    resolve: resolve,
                    reject:  reject
                };
            });
            promises.push(p);
            client.results = results;
            client.updates = updates;
            message.clientargs = clientArgs[idx];
            message.clientIdx = idx;
            idx++;
            client.obj.send(message);
        }
        return Promise.all(promises).then(() =>{
            for(let client in processes) {
                delete client.promise;
            }
            testfinished = true;
            //if (withMQ) {  }
            return updateResults(withMQ);
           
        }) .then(() => {
            clearInterval(txUpdateInter);
            return Promise.resolve();
        })
    }
    // launch clients
    processes = {};
    for(let i = 0 ; i < number ; i++) {
        launchClient(updates, results);
    }

    // start test
     return startTest(number, message, clientArgs, updates, results, withMQ);
}
module.exports.startTest = startTest;

/**
 * Send message to all child processes
 * @param {JSON} message message
 * @return {Number} number of child processes
 */
function sendMessage(message) {
    for(let pid in processes) {
        processes[pid].obj.send(message);
    }
    return processes.length;
}
module.exports.sendMessage = sendMessage;

/**
 * Stop all test clients(child processes)
 */
function stop() {
    for(let pid in processes) {
        processes[pid].obj.kill();
    }
    processes = {};
}
module.exports.stop = stop;

/**
 * Close kafka consumer
 */
function closeKafkaConsumer() {
      globalConsumer.close(() => {});
}
module.exports.closeKafkaConsumer = closeKafkaConsumer;
```

haojun:
@nitesh7sid 
I'm not sure to fully get your point. When you say the problem happens because "both the "if" and "else" block in the txUpdate function of local-client.js is executed", do you mean there are mq txns as well as non-mq txns in this test round?
I think the problem is caused by the newNum which is calculated for all txns (including those txns which have not included in the newResults array). One solution is to use txUpdated message to update the submitted value, and in verifyUnconfirmedTransactions(), set 0 to submitted property of dataToPush to avoid duplicate calculation. 

(Just an example, not sure it's 100% correct since i can not test it)

function txUpdate() {
......
let newStats;
let nonMqBuffer = [];
for (var i =0; i < newResults.length; i++){
     let txObject = newResults[i];
     if (txObject.GetneedVerifyWithMQFlag()){
           bufferToSend.push(txObject)
     }
     else {
           nonMqBuffer.push(txObject);
    }
}

// always send a txUpdated message to notify submitted count as well as non-mq results
if (nonMqBuffer.length === 0) {
    newStats = bc.createNullDefaultTxStats();
} else {
    newStats = blockchain.getDefaultTxStats(nonMqBuffer, false);
}
process.send({type: 'txUpdated', data: {submitted: newNum, committed: newStats}});

// send txUpdatedWithMq, the submitted attribute is only used to tell how many mq txns need to be check
if (bufferToSend.length !=0) {
    process.send({type: 'txUpdatedWithMQ', data: {submitted: bufferToSend.length, committed: bufferToSend}}); 
}

// apply trim, not sure how to apply trim to mq txns......
......
}

function verifyUnconfirmedTransactions(updates) {
   ......
   let newStats = blockchain.getDefaultTxStats(newResults, false);
   // the submitted value has already been counted, so set it as 0 here
   let dataToUpdate = {submitted: 0, committed: newStats};
   pushUpdate(global_pid, dataToUpdate);
}

Hope it answers your question : ）

BTW：I've handed over my caliper work to @panyu4 and @jiangyaoguo , they can help to review and merge this feature. I'm still happy to answer any questions if you need my help.

panyu4:
@nitesh7sid， when I run this code , I occurred this problem, the screen is continuing to print the same info " [Transaction Info] - Submitted: 1000 Succ: 0 Fail:0 Unfinished:1000". Do you know what's the matter ?
PS: I've already started the Kafka from another machine and change the listener-config.

panyu4:
@nitesh7sid， when I run this code , I occurred this problem, the screen is continuing to print the same info " [Transaction Info] - Submitted: 1000 Succ: 0 Fail:0 Unfinished:1000". Do you know what's the matter ?
PS: I've already started the Kafka from another machine and change the listener-config.

panyu4:
@nitesh7sid when I run this code, I  found my screen was continuing  printing the same info"[Transaction Info] - Submitted: 1000 Succ: 0 Failed:0 Unfinished:1000" after the submitted transaction was increased from 30 to 1000.  Do you know what's the matter . I've already started kafka from another machine.

panyu4:
@nitesh7sid ， when I run this code, I found my screen was continuing printing the same information "[Transaction Info] - Submitted: 1000 Succ: 0 Failed:0 Unfinished:1000" after the submitted transaction was increased from 30 to 1000.  Do you know what's the matter . I've already started kafka from another machine.

panyu4:
@nitesh7sid ， when I run this code, I found my screen was continuing printing the same information "[Transaction Info] - Submitted: 1000 Succ: 0 Failed:0 Unfinished:1000" after the submitted transaction was increased from 30 to 1000.  Do you know what's the matter . I've already started kafka from another machine.

panyu4:
@nitesh7sid ， when I run this code, I found my screen was continuing printing the same information "[Transaction Info] - Submitted: 1000 Succ: 0 Failed:0 Unfinished:1000" after the submitted transaction was increased from 30 to 1000.  Do you know what's the matter . I've already started kafka from another machine.

nitesh7sid:
@panyu4 this implies that event listener is not able to receive transactions and publish them into kafka queue. Could u pls check the below things:
1) the event listner configuration file https://github.com/nitesh7sid/caliper/blob/master/src/listener/listener-config.json with correct URL's
2) - KAFKA_ADVERTISED_HOST_NAME is set to the IP address of the machine running the kafka cluster
https://github.com/nitesh7sid/caliper/blob/master/network/kafka-setup/docker-compose-kafka.yaml

panyu4:
@nitesh7sid ，Thanks . It works now. I'm sorry for sending you so many repeated questions. That was happened just because of the shutting down of github I found  I was not able to comment it.
There are some problems that I think maybe you need to modify them.
1. the program  can not exit by itself when the test is finished. I have to manually exit with ctrl+c.
2. when I tried to run the test again, I have to remove the docker containers on the machine where the kafka is running, and then start the kafka again. After these have been done ,  I can run caliper in the withMQ mode.  I don't know if these are your test operations.  I think these operations are complex for people without professional skills.
3. I see there are some conflicting files. Would you please resolve them?

nitesh7sid:
Hi @panyu4 i tried the above code and seems to be working fine.
1) After the tests are complete, the program exists
2) After the tests are complete its a good practice to clear the MQ since the data already in MQ(transaction id's and confirmation time) will be irrelevant to the subsequent tests. Also time required to compute the statistics would be more if the MQ isn't cleared. Similar to how we clear the fabric ledger after the tests are done, users can also clear the MQ containers.
3) I have resolved merge conflicts. 

-Thanks

panyu4:
@nitesh7sid ,Thanks. 1 and 3 seems good to me.
2. ==> I mean if I do not clear the MQ container, the program will encounter an error to stop the tests. But  when I clear the MQ container first, it runs well. 
I list the operations below.
(1) run `docker-compose -f docker-compose-kafka.yaml up -d` on kafka machine
(2) run `node main.js -c config-with-mq.json`
(3) after the tests are completed , run `node main.js -c config-with-mq.json` again
The error Information got after the above operations is like this.
`TypeError: transactionObject.Set is not a function
    at Consumer.<anonymous> (/home/panyu/projects/caliper/src/comm/client/client-util.js:63:35)
    at emitOne (events.js:116:13)
    at Consumer.emit (events.js:211:7)
    at /home/panyu/projects/caliper/node_modules/kafka-node/lib/client.js:202:16
    at Object.<anonymous> (/home/panyu/projects/caliper/node_modules/kafka-node/lib/protocol/protocol.js:212:20)
    at Object.self.tap (/home/panyu/projects/caliper/node_modules/binary/index.js:248:12)
    at decodeMessageSet (/home/panyu/projects/caliper/node_modules/kafka-node/lib/protocol/protocol.js:174:8)
    at /home/panyu/projects/caliper/node_modules/kafka-node/lib/protocol/protocol.js:216:13
    at Gunzip.onEnd (zlib.js:131:5)
    at emitNone (events.js:111:20)`
`Error [ERR_IPC_CHANNEL_CLOSED]: channel closed
    at process.target.send (internal/child_process.js:578:16)
    at Timeout.txUpdate [as _onTimeout] (/home/panyu/projects/caliper/src/comm/client/local-client.js:66:13)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)`

I think it would be better to avoid this. Maybe we should clear the MQ Containers automatically and also should give some instruction to these errors like adding error information, adding instructions in the document

nitesh7sid:
Hi @panyu4 , thanks! I have handled the above error and it displays useful information for users to clear the MQ before running the tests.

panyu4:
@nitesh7sid  Thanks. 
It runs well for me. Now  when the program encounter the error, it will just show the information then exit. It may forget to remove the containers for the Fabric as what the original Caliper is done after completing all of the tests.  I think this step should be added.

nitesh7sid:
Hi Panyu, I have made the suggested changes and resolved linting issues.

panyu4:
@nitesh7sid , there are some conflicts in bench-flow.js.

src/comm/client/client-util.js
```diff
@@ -9,7 +10,80 @@
 'use strict';
 const logger = require('../util.js').getLogger('client-util.js');
 let processes  = {}; // {pid:{obj, promise}}
+let txUpdateTime = 1000;
+const kafka = require('kafka-node');
+const listener_config = require('../../listener/listener-config.json');
```
panyu4: Since Kafka mode is not necessary for all kinds of the tests,  these requirements may need to move into the related function.

nitesh7sid:
@panyu4 changes made as suggested. I am still not sure why the travis is failing. here are the logs of travis CI jobs https://travis-ci.org/hyperledger/caliper/jobs/456921827 and it looks like some fabric txns are failing.

src/fabric/e2eUtils.js
```diff
@@ -643,52 +644,53 @@ async function invokebycontext(context, id, version, args, timeout){
...
+        if (! withMQ) {
 
-                        // either explicit invalid event or valid event, verified in both cases by at least one peer
-                        invokeStatus.SetVerification(true);
-                        if (code !== 'VALID') {
-                            let err = new Error('Invalid transaction: ' + code);
-                            errFlag |= TxErrorEnum.BadEventNotificationError;
+            let newTimeout = timeout * 1000 - (Date.now() - startTime);
+            if(newTimeout < 10000) {
+                commUtils.log('WARNING: timeout is too small, default value is used instead');
```
panyu4: please rebase and use commLogger.warn('WARNING: timeout is too small, default value is used instead')  to record the log information. For info, commLogger.info('') can be used and commLogger.error('') is used for errors.

src/fabric/e2eUtils.js
```diff
@@ -709,17 +711,19 @@ async function invokebycontext(context, id, version, args, timeout){
...
-            commLogger.error('Failed to complete transaction [' + txId.substring(0, 5) + '...]: every eventhub connection closed');
-        } else {
-            invokeStatus.SetStatusSuccess();
+        if (!withMQ) {
+            await Promise.all(eventPromises);
+            // if the Tx is not verified at this point, then every eventhub connection failed (with resolve)
+            // so mark it failed but leave it not verified
+            if (!invokeStatus.IsVerified()) {
+                invokeStatus.SetStatusFail();
+                commUtils.log('Failed to complete transaction [' + txId.substring(0, 5) + '...]: every eventhub connection closed');
```
panyu4: commUtils.log should be replaced by commLogger.error

src/listener/block-listener-adapter.js
```diff
@@ -0,0 +1,70 @@
...
+* Copyright Persistent Systems 2018. All Rights Reserved.
+*
+* SPDX-License-Identifier: Apache-2.0
+*@file Implementation for block listener adapter.
+*/
+
+'use strict';
+const kafka = require('kafka-node');
+const Util = require('../comm/util');
+const log = Util.log;
```
panyu4: please use Util.getLogger() to log the information

src/listener/block-listener-adapter.js
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Create Kafka topic
+     */
+    createTopic() {
+        let self = this;
+        this.producer.on('ready', function () {
+            self.producer.createTopics([self.listener_config.topic], false, function (err, data) {
+                if (err) {
+                    log('Error creating Topic', err);
```
panyu4: log should be changed into util.getlogger

src/listener/listener-fabric.js
```diff
@@ -0,0 +1,112 @@
...
+*/
+
+'use strict';
+
+const Client = require('fabric-client');
+const fs = require('fs');
+const path = require('path');
+let testUtil = require('../fabric/util.js');
+const Util = require('../comm/util');
+const log = Util.log;
```
panyu4: replace Util.log with Util.getLogger()

src/listener/listener-fabric.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Fetch Blocks from Fabric peer and publish into kafka
+     *
+     */
+    getBlocks() {
+        let self = this;
+        self.client_kafka.on('error', function (error) {
+            log('Kafka client ERROR', error);
```
panyu4: log should be replaced with Util.getLogger()

src/listener/listener-fabric.js
```diff
@@ -0,0 +1,112 @@
...
+                    let payload = [{
+                        topic: self.listener_config.topic,
+                        messages: JSON.stringify(event_data),
+                        partition: 0,
+                        attributes: 1
+                    }];
+
+                    self.producer.send(payload, function (error, result) {
+                        if (error) {
+                            log('Error while publishing block in kafka', error);
```
panyu4: log should be replaced with Util.getLogger();  when logging errors with logger=Util.getLogger(), please use logger.error().

src/listener/listener-fabric.js
```diff
@@ -0,0 +1,112 @@
...
+                        }
+                    });
+
+                },
+                (err) => {
+                    log('Error in chaincode Event listener :', err);
```
panyu4: log should be replaced with Util.getLogger();  when logging errors with logger=Util.getLogger(), please use logger.error().

src/listener/listener-fabric.js
```diff
@@ -0,0 +1,112 @@
...
+                }
+                );
+            });
+
+        });
+        self.producer.on('error', function (error) {
+            log('Producer is not ready', error);
```
panyu4: log should be replaced with Util.getLogger();  when logging errors with logger=Util.getLogger(), please use logger.error().

src/comm/bench-flow.js
```diff
@@ -278,6 +280,8 @@ function defaultTest(args, clientArgs, final) {
  * @param {String} networkFile path of the blockchain configuration file
  */
 module.exports.run = function(configFile, networkFile) {
+    let localConfig = require(configFile);
+    configurationType = localConfig.test.clients.WITH_MQ;
```
panyu4: localConfig's value is same as config in line 290. configurationType  can be got after the config is set.

src/comm/client/client.js
```diff
@@ -90,14 +92,43 @@ class Client{
...
+    * @param {JSON} demo object which trakcs real time metrics
+    * @param {String} config path of the configuration file
+    * @param {String} absCaliperDir caliper directory
+    * @param {JSON} listener_child child process
+    * @param {JSON} t test object
     * @return {Promise} promise object
     */
-    init() {
+    init(demo, config, absCaliperDir, listener_child) {
+        absConfigFile = require(Util.resolvePath(config));
```
panyu4: since absConfigFile can be recorded using config framework, it is not necessary to add config as a parameter. 
You can use config framework like this:  const cfUtil = require('./config-util.js');  const cfUtil.getConfigSetting('fabric:WITH_MQ',false);   
If there is any problem when using this config framework, please let me know or you can contact with @feihujiang

src/comm/client/client.js
```diff
@@ -221,8 +258,9 @@ class Client{
      * @return {Promise} promise object
      */
     _startLocalTest(message, clientArgs) {
+
         message.totalClients = this.number;
-        return clientUtil.startTest(this.number, message, clientArgs, this.updates.data, this.results);
+        return clientUtil.startTest(this.number, message, clientArgs, this.updates.data, this.results, this.config.WITH_MQ);
```
panyu4: The last parameter is not necessary to be added since it can be got through the config framework.

panyu4:
@aklenik would you please review this pr?

panyu4:
@feihujiang @nklincoln would you please review it?

nitesh7sid:
As discussed with @panyu4 closing this PR since it got too large and raised new one here:
https://github.com/hyperledger/caliper/pull/236

Issue #95, opened 2018-06-14
Don't assume relative paths in config-files


RobertDiebels:
The fabric blockchain configuration file expects "relative-to-project-root" paths for certificates. Users should be able to pass absolute paths.

`path.join(__dirname, '../..', somepathvalue)`

I encountered this when creating a Docker container out of caliper.I mounted my configuration files into a different directory and was faced with a file not found exception even though I provided an absolute path to the files.

panyu4:
Since PR #133 had resolved this issue, I'm closing it.

PR #133, opened 2018-07-16
Use common utility for path resolution

aklenik:
Related to Issue #95, this PR unifies the path resolution of different modules, by using [the following](https://github.com/hyperledger/caliper/blob/6cf2891736c60dfd5c0a3221f36b1841dc79318d/src/comm/util.js#L51) utility function.

Now every path in the config files can be set either as a relative path to the Caliper root directory or as an absolute path.

@nklincoln I successfully run a Composer sample benchmark, but could you review the Composer-related part just in case?
@haojun Can you tag someone from the Iroha contributors to review the Iroha-related part? Although those were (I hope) straightforward modifications.

### DIFF
--- a/network/fabric/2-org-1-peer/composer-tls.json
+++ b/network/fabric/2-org-1-peer/composer-tls.json
@@ -1,6 +1,6 @@
 {  
   "composer": {
-    "chaincodes": [{"id": "basic-sample-network", "version": "0.1.0", "path": "contract/composer", "orgs": ["Org1", "Org2"], "loglevel": "INFO"}],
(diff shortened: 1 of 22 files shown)


haojun:
@stinger112 can help to review the iroha part

nklincoln:
works with Composer 🥇

haojun:
@stinger112 , would you please review it?

haojun:
@stinger112 I've tested it with iroha, it works according to my test. So i'd like to approve this pr first. If you find any problem later, you can open an issue on it.

PR #159, opened 2018-08-14
Added event subscription approach to get block commit status instead …

Ram-srini:
Added event subscription approach to get block commit status instead of rest get batch status query.

Signed-off-by: Ramakrishna Srinivasamurthy <ramakrishna.srinivasamurthy@intel.com>

### DIFF
--- a/src/sawtooth/sawtooth.js
+++ b/src/sawtooth/sawtooth.js
@@ -14,6 +14,152 @@ const log = require('../comm/util.js').log;
 let configPath;
 const request = require('request-promise');
 const TxStatus = require('../comm/transaction');
+const _ = require('lodash');
+const { Stream } = require('sawtooth-sdk/messaging/stream');
+const {
+    Message,
+    EventList,
+    EventSubscription,
+    ClientEventsSubscribeRequest,
+    ClientEventsSubscribeResponse,
+    ClientEventsUnsubscribeRequest,
+    ClientEventsUnsubscribeResponse
+} = require('sawtooth-sdk/protobuf');
+
+let lastKnownBlockId=null;
+let stream;
+let blockCommitSatus = new Map();
+let currentBlockNum=0;
+
+/**
+* Get the last recent block id for the block chain
+* @return {Promise<String>} last recent block id
+*/
+async function getCurrentBlockId() {
+    const request = require('request-promise');
+    let config = require(configPath);
+    let restApiUrl = config.sawtooth.network.restapi.url;
+    const blocks = restApiUrl + '/blocks?limit=1';
+    let options = {
+        uri: blocks
+    };
+    return request(options)
+        .then(function(body) {
+            let data = (JSON.parse(body)).data;
+            if (data.length > 0) {
+                currentBlockNum = parseInt(data[0].header.block_num);
+                lastKnownBlockId = data[0].header_signature.toString();
+                return currentBlockNum;
+            }
+        });
+}
+
+/**
+ * Get block data from event message
+ * @param {Object} events message
+ * @return {Promise<object>} The promise for the result of event message
+ */
+async function getBlock(events) {
+    const block = _.chain(events)
+        .find(e => e.eventType === 'sawtooth/block-commit')
+        .get('attributes')
+        .map(a => [a.key, a.value])
+        .fromPairs()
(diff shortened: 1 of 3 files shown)


Ram-srini:
@haojun @feihujiang - Please review changes for event subscription approach to get the block commit status instead of commit status query using rest-api.

haojun:
@jiangyaoguo @panyu4

Ram-srini:
@jiangyaoguo  @panyu4 @haojun  I have some improvements to event subscription code changes. I am doing one time event subscription in init() method and handling event every time it got the event. Please review changes.

Ram-srini:
@feihujiang  @haojun   - Please review this PR

src/sawtooth/sawtooth.js
```diff
@@ -14,6 +14,152 @@ const log = require('../comm/util.js').log;
...
+ * @param {Number} batchStats Batch status object to update commit status
+ * @return {Promise<object>} returns batch commit status
+ */
+function getBatchEventResponse(block_num, batchStats) {
+    return new Promise(resolve => {
+        while(blockCommitSatus.get(block_num) !== 'pending') {
+        }
+        //remove the block number from map because we are done with this block
+        blockCommitSatus.delete(block_num);
+        batchStats.SetStatusSuccess();
```
feihujiang: Why the next block is not pending, the batch commit status is success? Why the next block must include the batch?

src/sawtooth/sawtooth.js
```diff
@@ -14,6 +14,152 @@ const log = require('../comm/util.js').log;
...
+ * @param {Object} msg event message
+ * @return {void}
+ */
+async function handleEvent(msg) {
+    if (msg.messageType === Message.MessageType.CLIENT_EVENTS) {
+        const events = EventList.decode(msg.content).events;
+        getBlock(events).then(result => {
+            lastKnownBlockId = result.blockId.toString();
+            let blockNum=result.blockNum;
+            //On receiving event with block, update the status of the block to sucess
```
feihujiang: sucess is a typo.

src/sawtooth/sawtooth.js
Ram-srini: Once batch is submitted then the adding status of batch as pending to hash map and when we get the event then remove the entry from hash map and update the status to success and send it to caliper.

src/sawtooth/sawtooth.js
Ram-srini: After getting event from transactions commit event, updating the status to success.

src/sawtooth/sawtooth.js
feihujiang: Yes ,I know the total process. But Which event? I can't find the batch commit event, but the block commit event. So my question is why we get the  block commit event, we could update the status of the batch to success.

src/sawtooth/sawtooth.js
feihujiang: The last word 'sucess' of the annotation is a typo.

src/sawtooth/sawtooth.js
Ram-srini: We are subscribed to block-commit event. In sawtooth it will be block level, not batch level. Multiple batches are added to block and it is committed. After getting block commit event we are setting batch status to success and sending it to caliper.

src/sawtooth/sawtooth.js
Ram-srini: It is success

src/sawtooth/sawtooth.js
Ram-srini: Got your point, i am updating batch status to pending when submitted but if multiple batches goes to in single block then we are not updating the status of all batches, that is  question right?

Ram-srini: Yes that's correct. It should be fine because block consists of batches and each batch contain multiple transactions and block is ultimate entity getting added to block chain.

feihujiang: As the codes, when we get block(2) commit event, we are setting all of batches(a, b, c, d, e, f) status to success, but I think this is not correct.  The truth is that only (a, b, c) batches status is success. But the codes are setting all of the batches' status to success.

PR #161, opened 2018-08-16
Add a small feature for repeating test without reinstall chaincode

ski2per:
Signed-off-by: zhangpeng <ski2per@163.com>


ski2per:
Currently we are using Caliper as our performance test tool. Our environment is in Docker Swarm, I think this feature will be convenient for us to test same chaincode repeatly

Issue #142, opened 2018-07-25
BUG: some rate controllers are invalid


haojun:
Now an input argument 'results' is passed to rate controllers to help controllers learn the states of all finished txns. But since PR #122, 'results' don't contain all finished txns results, now it is a temparory array which only contains recent unhandled results. The change makes some rate controllers such as pid or composite rate controller invalid. Now a global variable 'resultStats' is used to save txns statistics (in resultStats[0], note the statistics does not include those in 'results' array), which can be useful for rate controller to get the neccessary information.

@aklenik @nklincoln

PR #164, opened 2018-08-20
fix PID rate controller for new client test mechanism

nklincoln:
Required fix for PID rate controller as part of #142 

-> change to use the current backlog directly
-> update unit tests to reflect change
-> add PID to the composer 'standard' test set config


feihujiang:
LGTM，Thanks.

PR #187, opened 2018-10-11
Fixed composite rate controller logic and docs.

aklenik:
Adjusted the composite rate controller to the new environment semantics as detailed by Issue #142.
The virtualization mechanism is not totally transparent now, but without passing more information to the rate controllers, this is the only way. Updated the docs accordingly.

@panyu4 since Nick has already fixed the PID controller, if this PR is accepted, then the issue can be closed.


aklenik:
Misinterpreted the meaning of the `results` array, will fix it soon.

panyu4:
I am accepting this PR  and  closing the Issue #142

PR #190, opened 2018-10-15
Changed the sawtooth adapter to take multiple rest end points

Ram-srini:
Signed-off-by: Ramakrishna Srinivasamurthy ramakrishna.srinivasamurthy@intel.com


Ram-srini:
Updated with with right sign off info..

Issue #189, opened 2018-10-12
Sawtooth: BatchBuilderFactory - Custom Batch Builders


mtrubs:
I was looking at the `sample` and `smallbank` sawtooth examples. I noticed that they have all their configuration in the `benchmark/*` directories. This is great and isolated. However, there is one file for each in `src/sawtooth/Application/*BatchBuilder.js`.

Is the intention to clone and modify the caliper project vs having a new project with caliper as an npm dependency? I do not see any other way to add a new `BatchBuilder` for specific use cases. I can happily add a `BatchBuilderFactory.register` or something to that end. But thought I would check first before implementing something and issuing a PR.

mtrubs:
here is roughly what i was thinking anyways.

the paths in all the files are still all wonk but it works.

https://github.com/mtrubs/caliper/commit/052e59a42216a43cbde207ff106cf41723914f35

feihujiang:
This was resolved, so closed it.

PR #191, opened 2018-10-15
Moved the batch builder factory configuration into the sawtooth.json configuration files

mtrubs:
https://github.com/hyperledger/caliper/issues/189

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,10 @@
   },
   "engine-strict": true,
   "engineStrict": true,
+  "files": [
+    "*",
+    "src/comm/template/report.html"
+  ],
   "dependencies": {
     "cbor": "^3.0.3",
     "commander": "^2.11.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -56,15 +60,11 @@
...
       "src/comm/client/**",
-      "src/comm/rate-control/pidRate.js",
-      "src/comm/rate-control/rateControl.js",
-      "src/comm/rate-control/rateInterface.js",
       "src/composer/**",
       "src/contract/**",
       "src/fabric/**",
       "src/gui/**",
       "src/iroha/**",
-      "src/sawtooth/**",
```
mtrubs: I could not tell if there was a reason for all of these excludes or not. I left most. The rate-control files above actually already have some coverage.

package.json
```diff
@@ -19,6 +19,10 @@
   },
   "engine-strict": true,
   "engineStrict": true,
+  "files": [
+    "*",
+    "src/comm/template/report.html"
+  ],
```
mtrubs: I needed to add this so that the template html file gets included on an `npm install`

hurf:
Thanks @mtrubs for the contribution. Can you squash the commits? And @feihujiang please take a look.

mtrubs:
@hurf squashed.

feihujiang:
LGTM, Thanks.

PR #196, opened 2018-10-18
Fix not exited normally

panyu4:
Signed-off-by: panyu4 <PY.panyu@huawei.com>


feihujiang:
Calling process.exit(), however, forces the process to exit before those additional writes to stdout can be performed. It is not a recommended practice.

PR #215, opened 2018-11-09
Add fixed feedback rate controller

feihujiang:
The fixed feedback rate controller will send input transactions at a fixed interval, but when too many transactions are unfinished, it will sleep a period of time.


feihujiang:
@aklenik I updated this PR as your advice, please take a look. And I'll document the rate controller in another PR.

feihujiang:
@aklenik I updated this PR according to your comments, please review again.

aklenik:
@feihujiang Merged it, now only the documentation is needed to the `gh-pages` branch.

Issue #140, opened 2018-07-20
Monitor disk r/w rate


pplam:
When performing a large transaction benchmark on `fabric`, I got very high `Traffic In/Out` for each `peer` and the specified `orderer` of `1.1 GB ~ 2.5 GB` from the statistic result but nothing about the disk read/write rate of these nodes. Is it possible to monitor the hard disk read/write rate of each node?

haojun:
@pplam I'm 100% not sure, but @jiangyaoguo we can investigate on it.

aklenik:
Resolved by PR #237

PR #237, opened 2018-11-27
Enable display of disc r/w

nklincoln:

Targets #140 

This PR enables hooking into the the disc r/w portion of dockerstats output for local tests, and updates the output tables in the console.

It was not needed to update the html file for test output as the template deals with the additional column.

I have noticed in the output that disc read remains at zero, but this corresponds to the actual dockerstats output during the test run

### DIFF
--- a/src/comm/monitor/monitor-process.js
+++ b/src/comm/monitor/monitor-process.js
@@ -11,7 +11,7 @@
 const ps = require('ps-node');
 const usage = require('pidusage');
 const MonitorInterface = require('./monitor-interface');
-const Util = require('./util.js');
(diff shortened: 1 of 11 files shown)


src/comm/monitor/monitor-process.js
```diff
@@ -105,6 +105,7 @@ function getUsage(pids, type) {
         });
 
         Promise.all(promises).then((stats) => {
+            console.log(stats);
```
aklenik: This logs the CPU and memory stats for every progress update (maybe a leftover debug log?)

PR #239, opened 2018-11-28
modify config file format from json to yaml 

houqinghui:
modify config file format from json to yaml 
1. modify the config file format from to yaml.
2. change the config parsing codes.
@panyu4 @aklenik


package.json
```diff
@@ -31,8 +30,14 @@
   "dependencies": {
     "cbor": "^3.0.3",
     "commander": "^2.11.0",
+    "composer-admin": "^0.19.0",
```
aklenik: `js-yaml` should be added to the dependencies, since it's a Caliper-level dep.

The platform-specific dependencies should not be saved to `package.json`:
* `composer-*`
* `fabric-*`
* `grpc`
* `protocol-buffers`
* `sawtooth-sdk`

houqinghui: ok, i will do it.

aklenik:
@houqinghui Please resolve the two conflicts (`config-composer.json` was probably modified by another PR, so modify the related yaml config, and you can delete the json version; the `monitor.js` file was moved into an other directory)

PR #269, opened 2018-12-24
add caliper FAQ

houqinghui:
add caliper FAQ.

### DIFF
--- a/docs/Caliper_FAQ.md
+++ b/docs/Caliper_FAQ.md
@@ -1,0 +1,62 @@
+---
+layout: page
+title:  "Caliper FAQ"
+categories: opensource
+---
+
+
+### I. Environment, Platform & Version
+**Q:** How do I run Caliper to test a blockchain system?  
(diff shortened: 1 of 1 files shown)


nklincoln:
Quite a few of the items here should probably point to the relevant section of the docs ... and if those sections of the docs are not clear enough then they should be modified/enhanced.

My suggestion would be that everything above (but not including) "How to test the blockchain system that caliper does not support currently" link to existing documentation, and everything else remain unchanged

houqinghui:
> Quite a few of the items here should probably point to the relevant section of the docs ... and if those sections of the docs are not clear enough then they should be modified/enhanced.
> 
> My suggestion would be that everything above (but not including) "How to test the blockchain system that caliper does not support currently" link to existing documentation, and everything else remain unchanged

what do you mean the existing documentation is ? should i  make a new documentation for that question?

houqinghui:
@nklincoln @aklenik  i have revise the FAQ documentation, please review it.

houqinghui:
@panyu4 @aklenik  I update the FAQ documentation, please review it again.

Issue #226, opened 2018-11-22
Caliper test reports showing inflated values for throughput


hariharan-avs:
## Summary
The throughput obtained from test is very high than the provided input.

## Context
The caliper test is run in Sawtooth Nightly (1.1) on AWS

## Expected Behavior
Input 70 tps throughput 70 or lesser tps 

## Actual Behavior
Input 76 tps : throughput 117 and shows fabric docker images which are not used for this test.

## Steps to Reproduce
1. Install Sawtooth 1.1 Nightly on VMs in AWS instance
2. Install caliper on AWS
3. start sawtooth and run caliper tests with respective sawtooth.json rest end points
4. obtain caliper report

[report20181119T092133.zip](https://github.com/hyperledger/caliper/files/2607535/report20181119T092133.zip)

houqinghui:
@hariharan-avs  do you solve your problem? or you can list the detailed log.

nklincoln:
looks to have been fixed in the above ... closing

PR #276, opened 2019-01-01
fix the tps calculation

houqinghui:
fix the tps calculation


aklenik:
@houqinghui Could you provide some description on what was the problem, and how the PR fixes it?

panyu4:
@aklenik @houqinghui This PR is trying to resolve the issue #226. But after a few tests, this computation method that minuses the idle intervals and only the actual time is used for successful transactions' lasted time computing, will result in this phenomena  the computed TPS would be often larger than  sending rate since when sending rate is lower, the backend network will have a lot of idle time. So I think this pr still needs to be improved.

PR #282, opened 2019-01-15
burrow query

gregdhill:

Provides the ability for the Burrow adaptor to query the installed contract.


gregdhill:
cc: @aklenik

PR #289, opened 2019-01-21
Fix file mode for Fabric not supporting sendSignedTransaction

feihujiang:

To implement file mode need Fabric supporting sendSignedTransaction function.


panyu4:
Ref #286

Issue #110, opened 2018-06-29
Improve caliper to support long time testing


haojun:
Now txns statuses are stored in internal arrays and exchanged between processes via messages or zookeeper. Some people may want to start a long time testing such as testing 24 hours with a sending rate of 10000 tps. That may cause performance issues for caliper itself. So the mechanism of txns statuses storing and exchanging should be improved to meet the needs of long timing testing.

nklincoln:
closing due to:
- ability to use Prometheus as a txn store
- no longer building an array of Promises to wait on

PR #291, opened 2019-01-24
Add large number tests

feihujiang:

### DIFF
--- a/src/adapters/fabric/e2eUtils.js
+++ b/src/adapters/fabric/e2eUtils.js
@@ -540,7 +540,7 @@ async function getcontext(channelConfig, clientIdx, txModeFile) {
 
         // an event listener can only register with the peer in its own org
         if(org === userOrg) {
-            let eh = client.newEventHub();
+            /*let eh = client.newEventHub();
             eh.setPeerAddr(
                 peerInfo.events,
                 {
@@ -551,7 +551,8 @@ async function getcontext(channelConfig, clientIdx, txModeFile) {
                     'grpc.keepalive_time_ms' : 360000   // time to wait for ping response, 6 minutes
                     // 'grpc.http2.keepalive_time' : 15
                 }
-            );
+            );*/
+            let eh = channel.newChannelEventHub(peer);
             eventhubs.push(eh);
         }
     }
@@ -579,7 +580,6 @@ module.exports.getcontext = getcontext;
  * @async
  */
 async function releasecontext(context) {
-
     if(context.hasOwnProperty('eventhubs')){
         for(let key in context.eventhubs) {
             const eventhub = context.eventhubs[key];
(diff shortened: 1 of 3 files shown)


feihujiang:
## Exist Issues
#110

src/adapters/fabric/e2eUtils.js
```diff
@@ -551,7 +551,8 @@ async function getcontext(channelConfig, clientIdx, txModeFile) {
                     'grpc.keepalive_time_ms' : 360000   // time to wait for ping response, 6 minutes
                     // 'grpc.http2.keepalive_time' : 15
                 }
-            );
+            );*/
+            let eh = channel.newChannelEventHub(peer);
```
aklenik: This call is allowed by the 1.1 SDK, but it's not compatible with a 1.0 Fabric backend.

src/adapters/fabric/e2eUtils.js
```diff
@@ -733,6 +733,8 @@ async function sendTransaction(context, signedTransaction, invokeStatus, startTi
             errFlag |= TxErrorEnum.OrdererResponseError;
             invokeStatus.SetFlag(errFlag);
             invokeStatus.SetErrMsg(TxErrorIndex.OrdererResponseError,err.toString());
+            invokeStatus.SetStatusFail();
```
aklenik: As stated by the comments above, an error/timeout from the broadcast call doesn't necessarily mean that the ordering failed. It's just an ACK from the orderer, that might get lost. That's why the events will determine the final state of the transaction.

src/adapters/fabric/e2eUtils.js
```diff
@@ -859,13 +861,16 @@ async function invokebycontext(context, id, version, args, timeout){
...
-                let err = new Error('Endorsement denied: ' + proposal_response.toString());
+                //let err = new Error('Endorsement denied: ' + proposal_response.toString());
                 errFlag |= TxErrorEnum.BadProposalResponseError;
                 invokeStatus.SetFlag(errFlag);
-                invokeStatus.SetErrMsg(TxErrorIndex.BadProposalResponseError, err.toString());
+                invokeStatus.SetErrMsg(TxErrorIndex.BadProposalResponseError, 'Endorsement denied: ' + proposal_response.toString());
                 // explicit rejection, early life-cycle termination, definitely failed
                 invokeStatus.SetVerification(true);
-                throw err;
+                commLogger.error('Failed to complete transaction [' + txId.substring(0, 5) + '...]:' + ' Endorsement denied: ' + proposal_response.toString());
```
aklenik: The exception is thrown "to the end", so the errors can be handled at one place at the end of the function. Otherwise, there will be a lot of repetitive code.

src/comm/rate-control/fixedFeedbackRate.js
```diff
@@ -62,9 +62,13 @@ class FixedFeedbackRateController extends RateInterface{
             return;
         }
 
-        let diff = (this.sleepTime * idx - ((Date.now() - this.total_sleep_time) - start));
+        /*let diff = (this.sleepTime * idx - ((Date.now() - this.total_sleep_time) - start));
         if( diff > 5) {
             return await util.sleep(diff);
+        }*/
+
+        if(idx % 100 === 0) {
```
aklenik: What's the reason for sleeping 5ms for every 100th transaction?

aklenik:
After consulting with @feihujiang, we decided to close this PR for now and reopen it at a later time

Issue #292, opened 2019-02-01
how to specify the endorser?


lenoAlex:
for example, using 2org2peer of network, how to specify one peer in each org to endorse a transaction?
In another word, there are four peers, but only need two peers to process endorsement. Could it be configurable on caliper?

houqinghui:
@lenoAlex  The endorsement-policy defines the endorsement policy for all the chaincodes. Currently, only one policy can be defined and the policy is applied to all chaincodes. You can refer to https://hyperledger.github.io/caliper/docs/Fabric_Configuration.html and https://hyperledger-fabric.readthedocs.io/en/release-1.3/endorsement-policies.html

feihujiang:
As the fabric documentation says "if no single peer is preferable based on the criteria, the fabric client SDK will randomly select from the peers that best meet the criteria", so Caliper now uses random peers ( one peer per organization ) as endorsers as default. Here is [the related fabric documentation's link](https://hyperledger-fabric.readthedocs.io/en/release-1.4/discovery-overview.html#how-service-discovery-works-in-fabric).

lenoAlex:
@houqinghui @feihujiang if caliper can use peer which is specified by users and configurable to endorse, that could be better, i think.

feihujiang:
@lenoAlex In Caliper latest version, target peers are chosen using `peers[clientIdx % peers.length]` which would cycle the client connections through the available peers . #255

lenoAlex:
@feihujiang as far as I known, caliper uses interface "sendTransactionProposal(request, timeout)" of SDK to send  request of endorsement. And the arg, "targets" , of the struct , "ChaincodeInvokeRequest",  specifies a peer to endorse.

aklenik:
@lenoAlex The `fabric` adapter in Caliper uses an automatic client process-based load balancing between the possible peers of an organization. You have the following options:

1. If you don't want a peer to endorse transactions, just leave it out of the configuration file. Con: it definitely won't endorse, since Caliper can't "see" it.
1. If you add a peer to the configuration, it will be part of the automatic load balancing between peers. Con: you can affect the targets.
1. Check out the new `fabric-ccp` adapter #296 that allows the user callback to specify the target endorsers explicitly on a per-transactions basis. If not specified, then a (configurable) client- or transaction-based load balancing is used.

lenoAlex:
@aklenik ， I got it, thank you.

Issue #227, opened 2018-11-26
Support for Fabric versions 1.3


lenoAlex:
I am currently working on a Fabric network test that are only available from Fabric v1.3.  Are there any tips or workarounds available? Are there any plans to support newer versions and if so, what is the current state?

Adjusting the versions of fabric-client and fabric-ca-client while using the correct docker images did not work.

aklenik:
@lenoAlex The support for Fabric 1.3 is on the way, sometimes this week (upper bound).

lenoAlex:
@lenoAlex Is there a working fork that I could test? I developed some features base on  Fabric v1.3 and want to get some benchmark.

luigima:
Does the release date for 1.3 support still hold @aklenik?

lenoAlex:
@aklenik Any news regarding the release date for 1.3 support?

luigima:
Reminds me of my issue some weeks ago :P #185

aklenik:
@luigima @lenoAlex The new adapter is basically finished, but I still have to add example benchmarks and networks, plus the documentation of course :) But it's good to know that everyone is anxious to try it, it's a good kind of pressure :P

panyu4:
@lenoAlex @luigima @aklenik currently, caliper is able to support fabric v1.3 by the pr #279 . Pr #296 has tried a new fabric adapter, and after it is merged, you can play with fabric v1.3 by this new adapter.

aklenik:
Both Fabric adapters now support newer Fabric releases

Issue #185, opened 2018-10-10
Support for Fabric versions above 1.1


luigima:
I am currently working on a Fabric network that leverages private data collections that are only available from Fabric v1.2+. Unfortunately I did not manage to get caliper working with versions above 1.1 yet. Are there any tips or workarounds available? Are there any plans to support newer versions and if so, what is the current state? 

Adjusting the versions of fabric-client and fabric-ca-client while using the correct docker images did not work. I get invalid endorsement system chaincode errors while instantiating.

aklenik:
@luigima I'm almost (~80%) finished with the support for newer versions (and features). As soon as I have a working code-base (with a minimal documentation), I'll drop you the link of the fork here, so you can help the reviewers test it :) Expect it in 2-3 days.

luigima:
This would be awesome! Thank you in advance!

jiawanggao:
@aklenik how to get the code for supporting Fabric v1.2?thanks!

aklenik:
@jiawanggao it's not integrated into Caliper (or a fork) yet, but as soon as I have a working fork, I'll drop the repo url here

hurf:
@aklenik Would create a PR and marked with WIP be better? People can disscuss here instead of looking for other repos.

aklenik:
@hurf Will do that, good idea, but currently, my Fabric adapter implementation is in a sandbox project, but I'll open a WIP PR as soon as it's integrated into my Caliper fork.

luigima:
@aklenik Any news regarding a working fork that I could test?

lichanan:
> @jiawanggao it's not integrated into Caliper (or a fork) yet, but as soon as I have a working fork, I'll drop the repo url here
where is the url?could you publish it here

aklenik:
New releases of Fabric are now supported by both adapter

Issue #139, opened 2018-07-19
Support multiple orderers


pplam:
For `fabric` benchmark, only 1 `orderer` can be specified in `fabric.json`. This make all traffic go to 1 `orderer` and the test result is not really correct. Could you please implement the feature that support multiple `orderer` sooner?

pplam:
Got it, so i need to do some hacking on `caliper` and keep looking forward to this feature. Thanks

aklenik:
If you need the feature right away, then yes :) But I'm working on the issue, hopefully, it will be ready soon.

pplam:
That's great, I've done my job with some other monitor tools and if the feature is ready, I may use it later. Thanks again.

cao0507:
> That's great, I've done my job with some other monitor tools and if the feature is ready, I may use it later. Thanks again.

What monitor tools have you find?

cao0507:
@aklenik Hello，does caliper support multiple orderers now?

aklenik:
@cao0507 There will be a work-in-progress PR soon (hopefully within a day or two), and you can start playing around with the new features.

samlee1203:
is it ready now? I am also looking for caliper's fabric testing on multiple orderers (not solo).

rohitkhatri:
Is there any workaround to test this on multiple orderer until It's added?

aklenik:
@samlee1203 @rohitkhatri Check out PR #296 and the corresponding documentation. It's not merged yet, and every feedback is welcome!

aklenik:
Resolved by PR #296

PR #296, opened 2019-02-04
Add connection profile-based Fabric adapter

aklenik:

## Issue/User story
Changes related to the adapter:
* Updated the sample benchmarks to also support the new adapter API
* Added settings for the new adapter in the default Caliper configuration file
* Added sample network configurations for the new adapter (from v1.0 to v1.2, including mutual TLS networks from v1.1). Most of the additions originate from these example configurations, which are easy to check/test
* Added lightweight Caliper-level dependecies (for general use, not Fabric-specific packages), like `tmp` and `version-compare`
* Changed some `require` calls to YAML parsing for reading network configuration files in `bench-flow.js`
* Integrated the new adapter into `blockchain.js`
* Added general utility functions for checking/asserting property and version-related information

## Directly related issues (and many more regarding the other adapter)
#37 #115 #139 #185 #227 #292
## Validation of the fix
For each sample benchmark, pick a configuration file (like the linear rate config), and run it with every network version/topology/`fabric-ccp*` configuration combination.

#297 

@hyperledger/caliper-committers @nklincoln Any help with the review if appreciated!


aklenik:
EDIT: I forgot to extend the `src/comm/transaction.js` class with the functions used by the adapter.

aklenik:
@panyu4 Fixed. Checked with the following commands (from the network dir):

`grep -Ril "goleveldb" . --include=fabric-ccp*.yaml`  
(only `*goleveldb`, `dev` and `customdomain` folders should be listed)

`grep -Ril "couchdb" . --include=fabric-ccp*.yaml` 
(only `*couchdb` folders should be listed)

benchmark/drm/publish.js
```diff
@@ -39,16 +46,38 @@ module.exports.run = function() {
...
         'item' : buf
     };
-    return bc.invokeSmartContract(contx, 'drm', 'v0', {verb : 'publish', item: JSON.stringify(item)}, 120)
-        .then((results)=>{
-            for (let i in results){
-                let stat = results[i];
-                if(stat.IsCommitted()) {
-                    ids.push(stat.GetResult().toString());
+
+    if (bc.bcType === 'fabric-ccp') {
```
feihujiang: Benchmark test modules should give transparent support to kinds of blockchain(different types or versions), so here it would be better to reduce the tests module's dependency with the certain type (or certain version) of blockchain system. Could you please move this Fabric-ccp dependency into the file [blockchain.js](https://github.com/hyperledger/caliper/blob/master/src/comm/blockchain.js) ? And for the parameter 'invoker', is it possible to extract a common invoker instance ? Then it could be used into kinds of blockchain system.  As for the parameter 'channel', can we  add one more parameter (eg. named as 'optionalPara') with data type Map(eg. `{fabric: {channel: 'mychannel'}}` ) to the function invokeSmartContract?  What do you think about this?

feihujiang:
@nklincoln , could you please review this pr and give some advice?

benchmark/drm/publish.js
aklenik: @feihujiang I thought a lot about this, and there is also TODO in the development doc about this issue (also highlighted in the [doc of the adapter](https://aklenik.github.io/caliper/docs/Fabric_Ccp_Configuration.html#the-run-callback)).

The problem is, that the abstract API of the adapters has an inconsistent signature compared to what is needed. My suggestion in the dev docs is that we adopt the Fabric SDK convention, meaning the following: the adapter functions (mainly invoke and query) take some `context` for the test round like in the current version; an `options` describing everything related to the call, and a `timeout`. 

The `options` is a platform-specific object since, for example, each platform requires a different way to identify a target smart contract (Fabric has a `channel->chaincode->version` "target ID", Sawtooth and Burrow has different ways). So the user callback can't really be platform independent, there will be a need for a "business logic -> platform syntax" translation if the user plans to target multiple platforms. 

The `if (bc.bcType === 'fabric-ccp')` line could be omitted by making the abstract adapter API consistent across platforms. But I like the idea of extracting some common concepts (like invoker) into the adapter interface. But we need to gather the common concepts of platforms first.

I'm not sure I understand the need for `{fabric: {channel: 'mychannel'}}`, could you give an example or use case for this parameter?

feihujiang: No, my suggestion isn't providing the abstract API of the adapters. I mean it's better that the specific benchmark test (simple/drm etc) could be transparent to the blockchain systems.
Yes, "business logic -> platform syntax" is necessarily needed. To each platform requires a different way to identify a target smart contract, Fabric has a `channel->chaincode->version` "target ID", Sawtooth and Burrow has different ways. Could we add a mapping from a common business logic name to the target smart contract on the network Configuration File? So that the benchmark test (simple/drm etc) could use the common business logic name. If we could, {fabric: {channel: 'mychannel'}} is meaningless, avoid it.

aklenik: @feihujiang That's exactly what I was thinking after I wrote my previous comment. Just like users are identified with a unique name in this adapter, a chaincode specification inside a channel could also be labeled with a `smartContractId` attribute, that must be unique on a network level. 

So the user callback could reference it as `targetContract: simple`, and the adapter would manage the `simple => mychannel/simple/v0` mapping (similar for the other adapters). The target nodes and the invoker identity could also be referenced by name; these are probably common concepts for each (private/permissioned) platform.

However, some attributes cannot be generalized, like the target orderer to use (this is optional, but the user should be able to define it if needed), the function to invoke inside the chaincode (okay, maybe this could be common), and the syntax of transaction arguments (array of strings, named parameters, transaction dependencies, batches, etc).

So a minimal platform-specific code/arguments cannot be avoided for a multi-target benchmark. But I think it's a good tradeoff, plus this code should be really short.

I could incorporate this contract mapping into this adapter, but what about the current signature of the `invokeSmartContract` function? The separate `contractID` and `contractVersion` parameters still bother me a bit :) The new ID name should also incorporate the version because it's easy to mess up the versions in the callback (`"1.0"` vs `"v1"`)

feihujiang: @aklenik Great job! It is a good way that the adapter would manage the `simple => mychannel/simple/v0` mapping (similar for the other adapters).
I agree with you that a minimal platform-specific code/arguments cannot be avoided for a multi-target benchmark. And I think trying our best to extract common concepts is benefit for users. 
The new ID name should also incorporate the version==>exactly.

aklenik: @feihujiang Then I'll add these features to this adapter, as a best-practice to follow for other adapters. So, for now, this adapter will follow the current API, but yeah, we'll need to consolidate it later 👍 
I'll ping you when it's ready.

feihujiang: :+1:  :)

aklenik:
@feihujiang I've added the contract ID feature, so the API for invoking a tx is the same as before (with default invoker value, so that also disappeared from the benchmarks). The only decision point is now for the arg of the invoke, which we can't really avoid for now. The older query API is a little rigid, I kept it consistent with the invoke API in this adapter.

Note: there was some conflict in package.json, so the PR now contains an additional merge commit that I can't seem to squash...

Issue #329, opened 2019-02-22
When fabric network is created, caliper don't need to execute command. 


clumsky:
## Context

For example,user use caliper to run benchmark/simple.
node main.js -n ../../network/fabric-v1.1/2org1peergoleveldb/fabric-go.json 
test run success.
But when use has own fabric network, user don't need to create default fabric network. Now the solution is user modify fabric-go.json to remove caliper.command.start and caliper.command.end.
## Expected Behavior

User  use node main.js -argv  -n ../../network/fabric-v1.1/2org1peergoleveldb/fabric-go.json  to start testing. The argv can control whether creating the default fabric network. 
## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-caliper)
- [ ] [GitHub Issues](https://github.com/hyperledger/caliper/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/caliper)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

aklenik:
@clumsky The start and end commands can contain arbitrary scripts, not just network starting/cleaning ones. But I agree with your point, that during development, it would be convenient to optionally skip the start and/or end scripts.

I'll write a separate issue about this soon, and you can edit your PR accordingly (because there is a nicer way for this setting :) )

PR #330, opened 2019-02-22
Providing a argv to control whether create default fabric network

clumsky:
Signed-off-by: TaoChen <clumsky@163.com>

## Issue/User story
#329 

after this PR
node main.js -n ../../network/fabric-v1.1/2org1peergoleveldb/fabric-go.json -d false
bench-flows.js will not call execAsync(command) to create default fabric network.


aklenik:
@clumsky 
1) Please make sure that your code passes the required linting rules. Currently, these are the errors:
```
/home/travis/build/hyperledger/caliper/benchmark/simple/main.js
  33:1   error  Missing JSDoc for parameter 'argv'   valid-jsdoc
  34:40  error  Trailing spaces not allowed          no-trailing-spaces
  37:13  error  Expected '===' and instead saw '=='  eqeqeq
  37:16  error  Strings must use singlequote         quotes
/home/travis/build/hyperledger/caliper/src/comm/bench-flow.js
  316:1  error  Missing JSDoc for parameter 'command_default'  valid-jsdoc
```

2) See Issue #331 for a detailed, clean an general solution for the problem. You can modify your PR accordingly.

benchmark/simple/main.js
```diff
@@ -57,7 +57,7 @@ function main() {
     let absNetworkFile;
     if(typeof networkFile === 'undefined') {
         try{
-            absNetworkFile = Util.resolvePath('network/fabric-v1.1/2org1peergoleveldb/fabric-go-tls.json');
+            absNetworkFile = Util.resolvePath('network/fabric-v1.1/2org1peergoleveldb/fabric-node-tls.json');
```
aklenik: Any reason for changing the default? Plus PR #287 will make the parameters mandatory, so there won't be any default values anyway.

src/comm/bench-flow.js
```diff
@@ -331,9 +335,10 @@ module.exports.run = async function(configFile, networkFile) {
     //let configObject = require(absConfigFile);
     let configObject = Util.parseYaml(absConfigFile);
     let networkObject = require(absNetworkFile);
+    let defaultObject = Util.parseYaml(absDefaultConfigFile);
```
aklenik: Instead of manually accessing the config file, please use the built-in config mechanism of Caliper. Because this way you could only change these settings through the config file (and we want to avoid changing files, that's the purpose of the PR). You can access the configs this way:

```js
// at the beginning of the file
const config = require('./config.util.js');

// ...
// in the run function
let skipStart = config.getConfigSetting('core:skipStartScript', false);
// because command line options are parsed as strings
skipStart = skipStart === true || skipStart === 'true'; 
let skipEnd = config.getConfigSetting('core:skipEndScript', false);
skipEnd = skipEnd === true || skipEnd === 'true';

// now you can use them in the two if conditions
```

clumsky:
@aklenik  This wey, we can use node  main.js --skipstart false(or true) --skipend false(or true) to set caliper whether starting and ending default fabric network. And if skipstart and skipend argv not set, default.yaml can control whether starting and ending default fabric network.
Finally，whether we should use this way to modify other benchmark test？

aklenik:
@clumsky The `main.js` files are just example bootstrap entry points for Caliper's `bench-flow.js/run` function. So they're not part of the Caliper code-base. Adding any features to those files is meaningless. 

Caliper has a hierarchical configuration mechanism, based on the `nconf` package, that provides a really flexible mechanism for setting the same option from different sources, just like the Fabric configuration mechanism. 

So if you add the `skipEndScript` and `skipStartScript` options to the `core` section of the config file (as default, fallback values), you can modify them in the following ways (in priority order):
1. During runtime: 
    ```js
    const config = require('relative/path/to/src/comm/config-utils.js');
    config.setConfigSetting('core:skipStartScript', true);
    config.setConfigSetting('core:skipEndScript', true);
    
    // reading the setting somewhere else will yield the true value, regardless of the deafult config file, environment and command line options
    ```
1. From the command line:
    ```sh
    node main.js -c config.yaml -n network.yaml --core:skipStartScript=true --core:skipStartScipt=true
    ```
    This will overwrite the default config file and environment options if provided

1. From environment variables (this might not work currently, but will be fixed soon):
    ```sh
    CORE_SKIPSTARTSCRIPT=TRUE node main.js -c config.yaml -n network.yaml
    ```
    This will overwrite the default config file options if provided

1. Change the values in the default config file

So this is the config mechanism of Caliper, and every runtime-related configuration (like skipping some feature) must be configured through this, so it will be accessible from every piece of code.

I'll add a detailed review soon about which lines you need or not.

benchmark/simple/main.js
```diff
@@ -37,6 +37,8 @@ function main() {
     program.version('0.1')
         .option('-c, --config <file>', 'config file of the benchmark, default is config.json', setConfig)
         .option('-n, --network <file>', 'config file of the blockchain system under test, if not provided, blockchain property in benchmark config is used', setNetwork)
+        .option('-s, --skipstart <argv>', 'argv of starting default fabric network, if not provided, default.yaml will provided')
+        .option('-e, --skipend <argv>', 'argv of ending default network, ,if not provided, default.yaml will provided')
```
aklenik: Line 40-41 is not needed, so this file won't be modified.

src/comm/bench-flow.js
```diff
@@ -20,6 +20,8 @@ const Report  = require('./report.js');
 const Client  = require('./client/client.js');
 const Util = require('./util.js');
 const logger = Util.getLogger('bench-flow.js');
+const config = require('./config-util.js');
+let program = require('commander');
```
aklenik: Line 24: commander is not needed in this file

src/comm/bench-flow.js
```diff
@@ -331,14 +333,21 @@ module.exports.run = async function(configFile, networkFile) {
     //let configObject = require(absConfigFile);
     let configObject = Util.parseYaml(absConfigFile);
     let networkObject = require(absNetworkFile);
+    config.setConfigSetting('core:skipStartScript', program.skipstart);
+    config.setConfigSetting('core:skipEndScript', program.skipend);
```
aklenik: Line 336-337: the set lines are not needed, `nconf` will set them automatically

src/comm/bench-flow.js
```diff
@@ -383,8 +392,9 @@ module.exports.run = async function(configFile, networkFile) {
             if (!networkObject.caliper.command.end.trim()) {
                 logger.error('End command is specified but it is empty');
             } else {
-
-                await execAsync(networkObject.caliper.command.end);
+                if(skipEnd){
```
aklenik: I think it should be `if (!skipEnd)`

src/comm/bench-flow.js
```diff
@@ -331,14 +333,21 @@ module.exports.run = async function(configFile, networkFile) {
...
+    skipEnd = skipEnd === true || skipEnd === 'true';
 
     try {
         if (networkObject.hasOwnProperty('caliper') && networkObject.caliper.hasOwnProperty('command') && networkObject.caliper.command.hasOwnProperty('start')) {
             if (!networkObject.caliper.command.start.trim()) {
                 throw new Error('Start command is specified but it is empty');
             }
-
-            await execAsync(networkObject.caliper.command.start);
+            if(skipStart){
```
aklenik: I think it should be `if (!skipStart)`

clumsky:
@aklenik  Now, we can use node main.js - --skipstart true(or false) --skipend true(or false). 
thanks for your suggestion.

src/comm/bench-flow.js
```diff
@@ -331,14 +332,21 @@ module.exports.run = async function(configFile, networkFile) {
     //let configObject = require(absConfigFile);
     let configObject = Util.parseYaml(absConfigFile);
     let networkObject = require(absNetworkFile);
+    config.setConfigSetting('core:skipStartScript', config.getConfig()._config.argv().get('skipstart'));
+    config.setConfigSetting('core:skipEndScript', config.getConfig()._config.argv().get('skipend'));
```
aklenik: Line 335-336 is confusing, basically, you could set the same setting with two different names. `core:skip(Start/End)Script` will be automatically set when the `config` object is included/created. So please remove these lines.

I know `core:skip(Start/End)Script` is longer than writing `skipStart/End` in the command line, but Caliper will support many configuration points for many modules, so it's critical to identify them hierarchically. 

Fabric also has some really long settings names, but it's intuitive and maintainable :)

src/comm/bench-flow.js
aklenik: This issue is still not resolved.

src/comm/bench-flow.js
aklenik: This issue is still not resolved

src/comm/bench-flow.js
clumsky: i got it

src/comm/config-util.js
```diff
@@ -38,8 +38,10 @@ module.exports.setConfigSetting = (name, value) => {
 //
 //Internal method to get an override setting to the configuration settings
 //
-exports.getConfigSetting = (name, default_value) => {
+module.exports.getConfigSetting = (name, default_value) => {
     const config = exports.getConfig();
+    config.set('core:skipStartScript', config._config.argv().get('skipStartScript'));
+    config.set('core:skipEndScript', config._config.argv().get('skipEndScript'));
```
aklenik: Please remove Line 43-44, since they're automatically done by nconf.

clumsky: you mean i should modify the function constructor() of Config class in src/comm/Config.js.  but it also need use set function. 
i am not understand about  "automatically done by nconf"

aklenik: When you execute the following:
`node main.js -c someconfig.yaml -n somenetwork.yaml --core:skipStartScript=false`
nconf will automatically save every argument (and environment variable, and attributes in the config file, etc.) when it's initialized. So you don't need to set `core:skipStartScript` manually. 

That's the beauty of nconf. If you need some new runtime options (like with this PR), you just: 
1. write the code for reading that setting
2. add a default to the config file
3. the user can set it from outside the program (like command line argument, env. variable)

So no need to "parse it" yourself.

clumsky:
@aklenik  Thanks for your help.  Is this the right way?

Issue #340, opened 2019-03-05
Modularize the rate controllers


aklenik:
## Context
Caliper uses rate controllers to signal the user callback modules when they can submit transactions. Currently, rate controllers have to provide/export a class containing [some functions](https://hyperledger.github.io/caliper/docs/Rate_Controllers.html#custom-controllers) to be usable by Caliper (like a constructor, an init, apply and end functions). 

Also, Caliper can only use rate controllers that are explicitly listed in the [rate controller gateway](https://github.com/hyperledger/caliper/blob/master/src/comm/rate-control/rateControl.js). This limits the easy extensibility of the mechanism.

## Expected Behavior
The proposal is to modularize the rate controller implementations in the following way: a module implementing a rate controller should only expose/export a factory function, like `createRateContoller(bc, opts):object`, that should return an object that has the aforementioned functions. This way the rate controller developer would have explicit control over the lifetime of a rate controller object. 

Moreover, external rate controllers also should be supported, i.e., the `src/comm/rate-control/rateControl.js` class should be able to resolve paths to external JS files, provided that they export the necessary factory function.

Corresponding documentation should also be provided that describes the general API of rate controllers.

## Possible Fix
The existing rate controller files (in the `src/comm/rate-control` directory) should be modified to export the aforementioned factory function instead of the rate controller class. 
Additionally, the `src/comm/rate-control/rateControl.js` constructor should be modified to create the rate controllers through the factory function. 
Also, if it encounters an unknown rate controller type that is a path to a JS file, it should try to access its factory function if it’s provided.

aklenik:
Resolved by PR #372

PR #372, opened 2019-03-14
Added factory functions to the rate controllers

aklenik:

Resolves enhancement issue #340 as suggested.
Also consolidates the API/implmenetations of existing rate controllers.


feihujiang:
LGTM but why rate controllers need msg.clientIdx, msg.roundIdx parameters?

aklenik:
@feihujiang The user module/callback factory functions will receive the same arguments. Since both are created through different factory functions, returning objects with different APIs, a single class could implement both a rate controller and a client module. So you can implement business state-based rate controllers, and with the client and round IDs, you can implement different behavior profiles for "different" client processes. So basically you can do any kind of workload modelling you want, which is nice :)

feihujiang:
Got it.

Issue #382, opened 2019-03-21
Report generation could be enhanced


nklincoln:
Current reports generated, whilst adequate, could be enhanced for readability

## Context
It would be nice if it was possible to provide a description to the test labels on the reports. It would also be more meaningful to link to test labels instead of test rounds in the LHS metadata links.

nklincoln:
Suggested fix in #383

aklenik:
Resolved.

PR #383, opened 2019-03-21
Modify report template to list based on test label

nklincoln:

Suggested fix for #382

Following the generation of some Caliper reports, I noticed that the report structure could be enhanced. I have changed the template and associated report code to follow the config file more closely and make the report more readable by:
- Group test rounds under each test label, with each round starting at zero for each label.
- Link to test labels instead of test rounds from the metadata side navigation (this is a more meaningful link)
- Enable the provision of a description for each test label that is taken from the config file.

I have attached a file that is generated by the code.

[report-20190320T153655.html.zip](https://github.com/hyperledger/caliper/files/2991612/report-20190320T153655.html.zip)

### DIFF
--- a/benchmark/simple/config.yaml
+++ b/benchmark/simple/config.yaml
@@ -8,6 +8,7 @@ test:
     number: 5
   rounds:
   - label: open
+    description: Test description for the opening of an account through the deployed chaincode
     txNumber:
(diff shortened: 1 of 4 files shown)


panyu4:
LGTM.

Issue #376, opened 2019-03-18
add the Caliper running environment detection script  


houqinghui:
When we run caliper to test our blockchain network, we should install many pre-requisity tools which is  fussy. So i want to write a static script to check whether they are installed, then install them if not. 
1. nodejs
2. nodegyp
3. docker
4. docker-compose
5. authority of linux user

aklenik:
@houqinghui Can you elaborate on the pre-reqs? You mean docker and other dependencies, or blockchain platform dependencies?

PR #397, opened 2019-04-01
add environment check script

houqinghui:
The script `env_check.sh`  can check whethe the pre-requisites of Caliper have been installed and gives the version. @panyu4 @aklenik 
#376


panyu4:
@houqinghui , good job. Here are some suggestions for this PR:
1.  keep the root folder clean: rebase your branch and move this script to the  folder  `packages/caliper-application/scripts`; 
2. when detecting a wrong version of software, it would be better to specify the right version and give some instructions of how to install a right version;
3. As for the script's name, I think the name 'pre-requisites-check-for-Linux.sh' would be better.

PR #399, opened 2019-04-01
Fix TPS computation for batch mode

panyu4:
Signed-off-by: panyu2 <PY.panyu@huawei.com>

## Issue/User story

When the  transactions are not sent evenly eg. sending several transactions with one workload, current TPS may give some strange throughtput since in some situations the time used for transactions is not accurately computed. 

## Existing issues
#226 

## Design of the fix

1. record the maximum final time of all successful and failed transactions;
2. modify the computation of successful transactions time to the all transactions used time 

## Validation of the fix

1. start Caliper normally with 'open' test module and 'txnPerBatch' set larger than TPS ;
2. open another terminal and stop the running blockchain network.


feihujiang:
LGTM

PR #408, opened 2019-04-15
Iroha adaptor

houqinghui:
@aklenik @panyu4 @feihujiang  Please reveiw the iroha adaptor.


panyu4:
@houqinghui ,  current iroha adapter works. Here are some suggestions which can be resolved by other PRs:  1. provide two peers network; 2. support iroha's batch mode; 3. separate current function `sendTransactions` into two functions that means one supports a single transaction to be sent and the other supports several transactions to be sent in one time.

PR #395, opened 2019-03-30
Fix conflicts between smallbank benchmark and fabric-ccp configuration 

quangtdn:
Signed-off-by: quangtdn <nmquang229@yahoo.com.vn>

This is just to fix the issue #394 . 

In the benchmarking methods smallbankOperations.js and query.js, for fabric-ccp type the test should convert transactions' arguments into Strings before sending the transactions.

PR #430, opened 2019-04-26
[updated] Fix conflicts between smallbank benchmark and fabric-ccp configuration

quangtdn:
Signed-off-by: quangtdn <nmquang229@yahoo.com.vn>

Please refer to #395  .
In the benchmarking methods smallbankOperations.js and query.js, for fabric-ccp type the test should convert transactions' arguments into Strings before sending the transactions.


panyu4:
Nice catch. It will be merged and the previous PR will be closed.

PR #462, opened 2019-05-28
Issue with transferring money in "simpletest.js"

foodman47:
### Issue
-  error when trying the transfer method for the first time

### Steps to Reproduce
- benchmark "benchmark/simple/transfer.js" in "benchmark/simple/config.yaml" withtout the fix
   => it will output a recurring error.

### DIFF
--- a/packages/caliper-application/benchmark/simple/transfer.js
+++ b/packages/caliper-application/benchmark/simple/transfer.js
@@ -1,0 +1,65 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/caliper-application/benchmark/simple/transfer.js
```diff
@@ -0,0 +1,47 @@
+/**
```
aklenik: Please use a license header like in https://github.com/hyperledger/caliper/blob/master/packages/caliper-fabric-ccp/lib/fabric.js

packages/caliper-application/benchmark/simple/transfer.js
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+        args = {
+            'verb': 'transfer',
+            'account1': account1,
+            'account2': account2,
+            'money': initMoney.toString()
+        };
+    }
+
+    return bc.invokeSmartContract(contx, 'account', 'v0', args, 10);
```
aklenik: account => simple
(The contract id is `simple` )

foodman47: Thank you for reviewing my code! I think now it should be all good.

aklenik:
@foodman47 Thanks for the quick fix and for your first contribution 👍

PR #485, opened 2019-06-19
Enable user to specify a client gateway test

nklincoln:

This PR extends the fabric-ccp adaptor to enable a user to specify the use of a Fabric Gateway based transaction by the addition of a `use-gateway: true` flag on the benchmark configuration file.

Changes:
- check/set of a `usegateway` flag within Fabric Network
- conditioning on fabric client version>=1.4.0 to use the new flag
- construction of required items (wallets and gateway contracts) during init/getContext phases
- fork in main submit/evaluate pathway to test with either the Gateway contract or the standard caliper client
- modification of integration test to use the alternative route (both client routes are tested)

Code flow:
- during the creation of clients, identities are conditionally added to an in-memory wallet
- during the getContext() method, a map-of-maps is created for each available client and all available chaincodes/smart-contracts: `Map<clientId, Map<chaincodeId, contract>>`
- if no `usegateway` flag is present, the standard transaction path is followed, otherwise, based on the invokingID and the chaincode identifier, a contract is extracted for the invokingID and then either `evaluateTransaction` or `submitTransaction` is called on the contract, with the passed test arguments

Documentation update is required:
- To use the new feature the flag `usegateway` must be added to the configuration file
- Using a gateway prevents the ability to produce detailed Caliper metrics of the transaction progress

Notes:
- Does not yet support discovery

### DIFF
--- a/packages/caliper-burrow/lib/burrowClientFactory.js
+++ b/packages/caliper-burrow/lib/burrowClientFactory.js
@@ -38,7 +38,7 @@ class BurrowClientFactory {
      * @returns {Object} the child process
      */
     async spawnWorker() {
-        const child = childProcess.fork(path.join(__dirname, './burrowClientWorker.js'));
+        const child = childProcess.fork(path.join(__dirname, './burrowClientWorker.js'), process.argv.slice(1), { env: process.env});
 
         const msg = {
             type: 'init',
--- a/packages/caliper-fabric-ccp/lib/fabricNetwork.js
+++ b/packages/caliper-fabric-ccp/lib/fabricNetwork.js
@@ -1275,6 +1275,7 @@ class FabricNetwork {
     isTlsEnabled() {
         return this.tls;
     }
+
 }
 
 module.exports = FabricNetwork;
--- a/packages/caliper-composer/lib/composerClientFactory.js
+++ b/packages/caliper-composer/lib/composerClientFactory.js
@@ -37,8 +37,8 @@ class ComposerClientFactory {
      * Spawn the worker and perform required init
      * @returns {Object} the child process
      */
-    async spawnWorker() {
-        const child = childProcess.fork(path.join(__dirname, './composerClientWorker.js'));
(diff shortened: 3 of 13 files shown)


aklenik:
@nklincoln A quick pre-review remark:
Since this only affects the client side (so both way can be used for the same 1.4.x backend network), the switch could be provided by a runtime setting (caliper-fabricccp-usegateway), so the same config file can be used for both mode

nklincoln:
Good point- and that provides even more flexibility

packages/caliper-fabric-ccp/lib/fabricNetwork.js
```diff
@@ -154,6 +155,8 @@ class FabricNetwork {
 
         this.mutualTls = CaliperUtils.checkProperty(this.network, 'mutual-tls') ? this.network['mutual-tls'] : false;
 
+        this.useGateway = CaliperUtils.checkProperty(this.network, 'use-gateway') ? this.network['use-gateway'] : false;
```
aklenik: Instead of the network config entry, something like this should be used in the `fabric.js` constructor:
https://github.com/hyperledger/caliper/blob/27ec5f94e70a4cb3856f2b1dfd26b3e2f95b087f/packages/caliper-fabric-ccp/lib/fabric.js#L166

packages/caliper-burrow/lib/burrowClientFactory.js
```diff
@@ -38,7 +38,7 @@ class BurrowClientFactory {
      * @returns {Object} the child process
      */
     async spawnWorker() {
-        const child = childProcess.fork(path.join(__dirname, './burrowClientWorker.js'));
+        const child = childProcess.fork(path.join(__dirname, './burrowClientWorker.js'), process.argv.slice, { env: process.env});
```
aklenik: Shouldn't the second argument be `process.argv.slice(1)`? The actual call to `slice` is missing. Same for the other client factories.

PR #497, opened 2019-07-02
Fixed expected parameter

GigaDev90:
## Checklist
 - [x]  How to recreate the problem without the fix
 - [x]  Design of the fix
 - [x]  How to prove that the fix works

## Issue/User story

When I try to use a remote docker and passing the link of it into the config file, like:
http://127.0.0.34:2375/all
Caliper is not able to fetch the stats data of all the docker containers.

I guess that a similar problem is present on the "else" of the same branch. line: 121 (NOT FIXED)
## Steps to Reproduce

1. To recreate the bug just start hyperledger caliper passing the endpoint to the remote docker.

> monitor:
>   type:
>   - docker
>   docker:
>     name: 
>         **- http://127.0.0.34:2375/all**
>   interval: 1;

 The result of the execution, without the fix, will be to do not be able to fetch the containers stats.

## Existing issues

- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/caliper)

## Design of the fix

In any case, the value inside filterName.remote[h].containers cannot exclude the '/'.
The value is coming from line: 49, the function url.parse is cutting the different parts of the url and "pathname" will include the slash.
Adding the slash will make it able to match the value contained inside filterName.remote[h].containers

Another possible solution could remove the slash in the beginning.
## Validation of the fix

I tried by my self and with this little fix everything works fine.

### DIFF
--- a/packages/caliper-core/lib/monitor/monitor-docker.js
+++ b/packages/caliper-core/lib/monitor/monitor-docker.js
@@ -109,7 +109,7 @@ function findContainers() {
                 return Promise.resolve();
             }
 
-            if(filterName.remote[h].containers.indexOf('all') !== -1) {
+            if(filterName.remote[h].containers.indexOf('/all') !== -1) {
                 for(let i = 0 ; i < size ; i++) {
                     let container = docker.getContainer(containers[i].Id);
                     this.containers.push({id: containers[i].Id, name: h + containers[i].Names[0], remote: container});


aklenik:
@GigaDev90 `/all` is more strict than `all`, if the condition is true for `/all` then it's also true for `all`. Are you sure this is what causes the problem?

GigaDev90:
I'm not sure about what you mean with strict. The function "Array.prototype.indexOf()" is working over index of an array:

**Definition and Usage**
The indexOf() method searches the array for the specified item, and returns its position.
The search will start at the specified position, or at the beginning if no start position is specified, and end the search at the end of the array.

Returns -1 if the item is not found.

[https://www.w3schools.com/jsref/jsref_indexof_array.asp](https://www.w3schools.com/jsref/jsref_indexof_array.asp)

So, it's not about matching the index of the string, instead is the index of the array.

aklenik:
@GigaDev90 My mistake, I thought `containers` was a string, but it is a string array. And `url.parse` leaves the leading `/` in the pathname, but the code tries to find it without `/`, all clear know :) Thanks for the fix 👍

PR #531, opened 2019-07-26
License check process add and fix missing licenses

Lynn14m:
Signed-off-by: lmuswere <lynn14m@gmail.com>

Added license check process. Missing licenses fixed/ incorrectly formatted licenses fixed to comply with license check process.

closes #531

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -20,7 +20,7 @@ set -o pipefail
 # Bootstrap the project
 npm run bootstrap
 
-# Run linting and unit tests
+# Run linting, license check and unit tests
 npm test
 
 echo "---- Publishing packages locally"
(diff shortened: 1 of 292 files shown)


aklenik:
@Lynn14m the Fabric-ccp adapter has been renamed just now, so there're some conflicts. Hopefully git recognizes the renames so a simple rebase&merge can solve the problem.

nklincoln:
Hi, thanks for this - you have no idea how happy this PR makes me!! One thought though: it is possible to run the license checker as a pre-test in each package, which will tidy up the licCheck in the root directory. For instance:

root/package.json checks parent level and ignores 'packages | node_modules'
root/packages/adaptor/package.js checkes itself and ignores 'node_modules'

The `lerna run test` will iterate through all the sub-package tests, and so it will run each license check for each package 👍

Issue #553, opened 2019-08-20
Enable Fabric benchmarking with a FileSystemWallet


nklincoln:
Currently the configuration of a target is a little tricky, in particular the specification of the required crypto material.

This could be made much easier by enabling the use of a file wallet to specify the required identities (with the associated crypto material) and either using the file wallet direct (if using a gateway trnsaction) or by extracting the identity from the wallet and creating the required profile from the contained materials

nklincoln:
closing as complete 👌

PR #555, opened 2019-08-20
Update docs for using a file wallet

nklincoln:

Associated docs for PR #554 and issue #553


docs/Fabric_Configuration.md
```diff
@@ -407,6 +416,25 @@ clients:
       # properties of the client
 ```
 
+### With File System Wallet
+If using a file system wallet to specify identies, each client name *must* correspond to one of the identies within the provided wallet. If you wish to specify an `Admin Client` then the naming convention is that of `admin.<orgname>`. If no explicit admin client is provided for an organisation, it is assumed that the first listed client for that organisation is associated with an administrative identity.
+
+When a file system wallet is specified, the only required property for the each client organisation name to which the client belongs.
```
aklenik: => the only required property for each client is the organisation name

Issue #535, opened 2019-07-30
Enable Caliper lifecyle to be controlled more via config file


nklincoln:
At the moment the lifecycle of a caliper test cannot be easily controlled through the test config file.

It would be really nice if you could conditionally enable/disable the following via the config file:

- [x] start/end commands (noStart/noEnd)
- [ ] init (noConfigure)
- [ ] install smart contracts (noDeploy)
- [ ] only perform the init phase (initOnly)

With these additional flags, it would be nice if `caliper-flow` was modified to use them at the same time being converted into an actual Class for easier consumption by  the CLI (or any other user).

For instance, it might be that we create a new `CaliperEngine.js` class (or similar) that is interacted with by the caliper flow.js script to perform the appropriate flow control on the object

aklenik:
The first item is already implemented: --caliper-skipstart/endscript

nklincoln:
I think it would be nice for a user to be able to use either command line flags, or explicit config settings in a test config file. 

For instance, in addition to picking up any command line flags, we could introduce an optional section in the test config file under the key `flow`:

```
---
flow:
    skipStart: true/false
    skipInit: true/false
    skipInstall: true/false
    skipEnd: true/false
    testOnly: true/false
test:
  ~
monitor:
  ~
```

Then in `caliper-flow.js`, we could condition each key stage with 

```
  if (skipStart || (configObject.flow && configObject.flow.skipStart))  {
      logger.info('Skipping start commands due to passed flag');
  } else {
      doTheThing()
  }
```

and any form of 'testOnly/initOnly' can be used to define/override the appropriate 'skip' flags

aklenik:
Closed by #556

PR #556, opened 2019-08-20
Condition Caliper flow

nklincoln:
Related to #535 and enables conditional caliper-flow actions of:
- run start scripts
- run init phase
- run install smart contract phase
- run test phase
- run end script phase

All the above controlled via a config item passed at the command line level

### DIFF
--- a/packages/caliper-core/lib/caliper-flow.js
+++ b/packages/caliper-core/lib/caliper-flow.js
@@ -57,21 +57,36 @@ module.exports.run = async function(absConfigFile, absNetworkFile, admin, client
     let skipEnd = Config.get(Config.keys.SkipEndScript, false);
 
     try {
-        if (networkObject.hasOwnProperty('caliper') && networkObject.caliper.hasOwnProperty('command') && networkObject.caliper.command.hasOwnProperty('start')) {
-            if (!networkObject.caliper.command.start.trim()) {
-                throw new Error('Start command is specified but it is empty');
-            }
-            if(!skipStart){
-                const cmd = 'cd ' + workspace + ';' + networkObject.caliper.command.start;
-                await CaliperUtils.execAsync(cmd);
+
+        // Conditional running of 'start' commands
+        if (skipStart || (configObject.flow && configObject.flow.skipStart))  {
+            logger.info('Skipping start commands due to passed flag');
+        } else {
+            if (networkObject.hasOwnProperty('caliper') && networkObject.caliper.hasOwnProperty('command') && networkObject.caliper.command.hasOwnProperty('start')) {
+                if (!networkObject.caliper.command.start.trim()) {
+                    throw new Error('Start command is specified but it is empty');
+                } else {
+                    const cmd = 'cd ' + workspace + ';' + networkObject.caliper.command.start;
+                    await CaliperUtils.execAsync(cmd);
+                }
             }
         }
 
-        await adminClient.init();
-        await adminClient.installSmartContract();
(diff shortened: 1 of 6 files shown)


packages/caliper-core/lib/caliper-flow.js
```diff
@@ -42,6 +42,65 @@ module.exports.run = async function(absConfigFile, absNetworkFile, admin, client
...
+        flowFlags++;
+    } else {
+        performStart = true;
+        performInit = true;
+        performInstall = true;
+        performTest = true;
+        performEnd = true;
+    }
+
+    if (flowFlags > 1) {
```
aklenik: I think will always be false since the `if-else` structure imposes precedence on the flags, and only one will be processed.

packages/caliper-core/lib/caliper-flow.js
```diff
@@ -53,79 +112,99 @@ module.exports.run = async function(absConfigFile, absNetworkFile, admin, client
...
+                }
             }
         }
 
-        await adminClient.init();
-        await adminClient.installSmartContract();
-        let numberOfClients = await clientOrchestrator.init();
-        let clientArgs = await adminClient.prepareClients(numberOfClients);
+        // Conditional network initialisation
+        if (!performInit) {
```
aklenik: Init also includes the creation of connection profiles, for example (or other general init steps for the adapter objects). Can we really skip this part?

packages/caliper-core/lib/config/config-util.js
```diff
@@ -59,9 +59,14 @@ const keys = {
     ZooAddress: 'caliper-zooaddress',
     ZooConfig: 'caliper-zooconfig',
     TxUpdateTime: 'caliper-txupdatetime',
-    SkipStartScript: 'caliper-skipstartscript',
-    SkipEndScript: 'caliper-skipendscript',
     Logging: 'caliper-logging',
+    Flow: {
+        StartOnly: 'start-only',
```
aklenik: 1. Should prefix these with `caliper-` to follow the convention. 
2. The key hierarchy results in the following names: `caliper-flow-start-only`, etc. (So, missing `flow-` prefix)

packages/caliper-fabric/lib/fabric.js
```diff
@@ -2279,6 +2280,13 @@ class Fabric extends BlockchainInterface {
      * @async
      */
     async installSmartContract() {
+        // With flow conditioning, this phase is conditionally required
+        if (!this.initPhaseCompleted ) {
```
aklenik: What if both init and install is skipped? Will the users be initialized?

packages/caliper-samples/benchmark/marbles/config.yaml
```diff
@@ -13,6 +13,11 @@
 #
 
 ---
+flow:
```
aklenik: Shouldn't this be in the `default.yaml` config file? That's where the `Config` class reads it from (as a last resort).
Plus these names don't correspond to the read configs (start-only, etc.). I prefer the skip-x approach since those can be combined in any way.

packages/caliper-fabric/lib/fabric.js
nklincoln: The next interaction with a SUT comes at the client spawn phase, which includes the configuration user credentials to interact .. so we are safe there.

packages/caliper-core/lib/caliper-flow.js
nklincoln: 😂  - good point!!

PR #566, opened 2019-08-30
restrict wallet use

nklincoln:

If performing a "caliper-flow-only-test" interaction, when using a file wallet that has been created without admin identities, the current implementation will fail, despite not needing those.

This PR conditions the fabric.js implementation surrounding wallets to:
- only look for locally stored information if *not* using a file wallet
- only build contract maps for identities listed in the wallet

The following scenarios are available:
- Perform a benchmark (standard transaction implementation)
- Perform a benchmark through a fabric gateway
- Use a file wallet to initialise users a benchmark (standard transaction implementation)
- Use a file wallet to initialise users a benchmark through a gateway
- Use file wallet to initialise non-admin users for benchmarking on a configured network

### DIFF
--- a/packages/caliper-fabric/lib/fabric.js
+++ b/packages/caliper-fabric/lib/fabric.js
@@ -15,13 +15,12 @@
 'use strict';
 
 const FabricClient = require('fabric-client');
+let FabricNetworkAPI = require('fabric-network');
 const {google, common} = require('fabric-protos');
 const {BlockchainInterface, CaliperUtils, TxStatus, Version, ConfigUtil} = require('caliper-core');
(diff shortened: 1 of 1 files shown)


packages/caliper-fabric/lib/fabric.js
```diff
@@ -15,13 +15,12 @@
 'use strict';
 
 const FabricClient = require('fabric-client');
+let FabricNetworkAPI = require('fabric-network');
```
aklenik: We shouldn't forget to add `fabric-network` as a dependency for all SUT bindings, otherwise, this will fail runtime.

PR #588, opened 2019-09-16
Integrate Caliper With Prometheus 

nklincoln:
Big set of changes here to enable:
- Prometheus monitor with generic user queries
- Prometheus observer for console displayed txn stats

Changes include:
- modify local client to conditionally push to a prometheus push gateway
- modify returned results to include start/end times of client test
- add new prometheus monitor
- tidy existing Docker monitor
- remove demo.js and move to a test observer class
- add test observer interface
- add prometheus based test observer
- rename monitor methods/variables to make them related to what the method/variable is used for
- add new docker-compose file that includes Prometheus/Grafana
- delete zoo related items (clients and CLI)
- modify report builder and report template to enable generic results to be itemised
- at a txReset message to local observer so the observers behave the same way


packages/caliper-core/lib/prometheus/prometheus-query-client.js
```diff
@@ -0,0 +1,166 @@
...
+    constructor(prometheusUrl) {
+        this.prometheusUrl = prometheusUrl;
+    }
+
+    /**
+     * Configure the target for the push
+     * @param {String} type query type
+     * @param {String} query the unique query append
+     */
+    async configureTarget(type, query) {
```
aklenik: Is `async` needed?

packages/caliper-core/lib/prometheus/prometheus-query-client.js
```diff
@@ -0,0 +1,166 @@
...
+        const target = url.resolve(this.prometheusUrl, type + query);
+        this.requestParams = url.parse(target);
```
aklenik: what if queries (due to late answers) overlap? The `this.X` variables will be overwritten. 
The options of a query could be an independent object, passed around, instead of saved into an instance variable.

packages/caliper-core/lib/test-observers/local-observer.js
```diff
@@ -0,0 +1,265 @@
...
+        if (this.testData.latency.max.length > this.testData.maxlen) {
+            this.testData.latency.max.shift();
+            this.testData.latency.min.shift();
+            this.testData.latency.avg.shift();
+            this.testData.latency.x.shift();
+        }
+    }
+
+    /**
+     *
```
aklenik: Missing docs

packages/caliper-core/lib/test-observers/null-observer.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+class NullObserver extends TestObserverInterface {
+
+    /**
+     * Constructor
+     * @param {String} configPath path of the configuration file
+     */
+    constructor(configPath) {
+        super(configPath);
+        Logger.info('Configured monitor');
```
aklenik: Monitor -> observer?

packages/caliper-core/lib/test-observers/test-observer.js
```diff
@@ -0,0 +1,89 @@
...
+
+        this.observer = factoryFunction(absConfigFile);
+    }
+
+    /**
+     * Perform an update
+     * @async
+     */
+    async update() {
+        this.observer.update();
```
aklenik: Shouldn't this `await` for the call, or return the Promise, so the caller can await on it?

packages/caliper-core/lib/test-observers/test-observer.js
```diff
@@ -0,0 +1,89 @@
...
+    startWatch(clientOrchestrator) {
+        this.observer.startWatch(clientOrchestrator);
+    }
+
+    /**
+     * Stop watching the test output from the orchestrator
+     * @async
+     */
+    async stopWatch() {
+        this.observer.stopWatch();
```
aklenik: Shouldn't this `await` for the call, or return the Promise, so the caller can await on it?

packages/caliper-core/lib/prometheus/prometheus-query-client.js
nklincoln: good spot, this could be converted to a `retrieveTargetParameters` method that is passed through 👍

nklincoln:
@aklenik - thanks for the review, I've made the suggested edits, in particular modified the manner in which queries are made to prevent overwriting of parameters - happy to do the same in the 
push client too (which had a similar pattern)

Issue #467, opened 2019-05-30
Clean up the Fabric CCP adapter


aklenik:
## Context
The adapter currently relies on the `FabricNetwork` module for network information which is intended for the user modules and breaks the code flow (and code completion support) in the adapter.

## Possible Fix
The adapter should process the network configuration file directly, treating the sections as self-contained objects.

## Progress of the refactoring
- [ ] Extract the network validation into a separate util module
- [ ] Separate the user/admin/registrar initialization from the `FabricNetwork` module
- [ ] Separate the channel creation from the `FabricNetwork` module
- [ ] Separate the channel joining from the `FabricNetwork` module
- [ ] Separate the chaincode installation from the `FabricNetwork` module
- [ ] Separate the chaincode instantiation from the `FabricNetwork` module

nklincoln:
Superseded by #940

PR #595, opened 2019-09-30
Add declarative validation for Fabric network config files

aklenik:

Resolves part of issue #467 
(Also moves the config runtime key names, so the Config.js module can use them without a circular dep issue.)

## Design of the fix
The validation part of the `fabricNetwork.js` module was extracted into `configValidator.js`, which uses the joi package to define and validate object schemas declaratively. This makes it easy to adapt the validation rule for future/changed schemas.

## Validation of the fix
Unit tests cover the entire schema, property-by-property (since the schema is declarative, code coverage is not representative). The expected error/joi messages are checked to ensure that the test targets the intended property error (and isn't masked by an other error).

Even though this is just an under-the-hood change, the Fabric adapter docs is also updated to make it more readable and consistent with the validation rules. Live preview: https://aklenik.github.io/caliper/vNext/fabric-config/

### DIFF
--- a/packages/caliper-core/lib/config/Config.js
+++ b/packages/caliper-core/lib/config/Config.js
@@ -20,6 +20,65 @@ const nconf = require('nconf');
 
 nconf.formats.yaml = require('nconf-yaml');
 
+const keys = {
+    Bind: {
+        Sut: 'caliper-bind-sut',
+        Sdk: 'caliper-bind-sdk',
+        Args: 'caliper-bind-args',
+        Cwd: 'caliper-bind-cwd'
+    },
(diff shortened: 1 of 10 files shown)


aklenik:
> +bazillion for the comprehensive testing with the code 👍
> 
> Might be worth using the same joi based validation for the config file too?

Yes, that's also what I had in mind. Plus, Joi schemas are "compiled" and can be reused. 
So this means we could also use them for basic stuffs, like input validation in different functions, in a self-documenting way. This could definitely improve the readability for cold paths. 
(But who knows, maybe it's fast enough also for hot paths)

Issue #569, opened 2019-09-03
Remove sample benchmarks folder


nklincoln:
With the new caliper-benchmarks repository in existence, we no longer need (and should not maintain) the samples within the main code.

These should be deleted, and `test-samples` added into the integration tests folder that are used as self contained integration tests.

In addition, we should highlight the existence of the new benchmarks repository! 

Required tasks:
- modify integration tests to use directed networks specified and maintained within the `caliper-tests-integration` directory
- delete the `caliper-samples` directory
- Add caliper-benchmarks directory links and description to docs ... and make it super obvious!!

PR #605, opened 2019-10-03
Add multi-phase Fabric CI test

aklenik:

Adds a multi-phase Fabric test for CI, that is independent of the samples (because of #569)

## Specifics
SUT details (`packages/caliper-tests-integration/fabric-tests` dir):
* 2 Orgs, 1-1 peer, 1-1 CA, 2 Raft orderers, mutual TLS
* 2 channels, Marbles golang CC (with rich query) in both channel
* Benchmark configs:
    * 3 rounds (init, init, query)
    * With 3 different rate controllers
    * Different observer and monitor combination for each phases
* Dedicated network config for all phases, stripped to the required minimum
* Phases
    1. Only start the network
    2. Only initialize the network
    3. Only deploy the contracts (channels already created)
    4. The previous phase again to check the correct detection of already deployed contracts
    5. Only performing the testing phase through the low-level SDK API
    6. Only performing the testing phase through the gateway SDK API
    7. Only disposing of the network

## Remarks/Decisions
1. CAs can be removed since no dynamic user registration occurs
2. Or could add dynamic user registration

### DIFF
--- a/packages/caliper-tests-integration/.dockerignore
+++ b/packages/caliper-tests-integration/.dockerignore
@@ -16,5 +16,6 @@
 scripts/*
 node_modules/*
 log/*
+fabric_tests/*
 caliper.Dockerfile
 package.json
--- a/packages/caliper-tests-integration/fabric_tests/.gitignore
+++ b/packages/caliper-tests-integration/fabric_tests/.gitignore
@@ -1,0 +1,2 @@
+**/*.log
+report.html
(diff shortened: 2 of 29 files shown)


aklenik:
@nklincoln Updated PR with:
1. Rebased to caliper/master
2. Included crypto-gen step into the CI script
3. Overridden some default Caliper settings (with the new logging configs) for the test

PR #608, opened 2019-10-07
Update log formatting docs

aklenik:

Docs for PR #607


aklenik:
> careful - you are getting dangerously close to becoming a docs writer!!

Well, sometimes you need some distance from coding :D

PR #639, opened 2019-11-05
Flatten round configuration and consolidate related components

### LINKED ISSUE #636: Flatten the round descriptions in the benchmark configuration
## Context
The round descriptions in the benchmark configuration file use an array notation for some attributes as shorthand for reusing `callback` and `arguments` attributes values.
YAML anchors can achieve the same, with the following benefits:
1. The reuse is explicit and more flexible (anchors can be used anywhere, making every part of the config file "extendable").
2. The resolving of anchors happens when the YAML is loaded, so the core round processing code can be simplified.

## Expected Behavior
The round descriptions don't allow using array notation.

## Actual Behavior
The round descriptions allow using array notation.

## Possible Fix
Take the following config, for example:
```yaml
test:
    clients:
        type: local
        number: 1
    rounds:
    - label: init
      txNumber: [100, 200, 300]
      rateControl: [{ type: 'fixed-rate', opts: { tps: 20 } }, { type: 'fixed-rate', opts: { tps: 40 } }, { type: 'fixed-rate', opts: { tps: 60 } }]
      callback: ../init.js
      arguments:
          arg1: 100
          arg2: text
          arg3: text2
```

The same with YAML anchors and aliases would look like this:
```yaml
definitions:
    initRound: &init-round
        label: init
        txNumber: 100
        rateControl:
            type: fixed-rate
            opts:
                tps: 20
        callback: ../init.js
        arguments:
            arg1: 100
            arg2: text
            arg3: text2
test:
    clients:
        type: local
        number: 2
    rounds:
    - <<: *init-round
    - <<: *init-round
      txNumber: 200
      rateControl:
          opts:
              tps: 40
    - <<: *init-round
      txNumber: 300
      rateControl:
          opts:
              tps: 60
```

A few things to note:
1. Configuration files could be decomposed any way the user wants, Caliper doesn't enforce any (unnecessary) structure.
2. The differences between the rounds are more evident.

## Your Environment
* Version used: 0.2.0

aklenik:
Resolves #636 

The PR includes:
- [x] Converting `caliper-flow` to a class, and renaming it to `CaliperEngine`
- [x] Flattening the round configuration syntax, and moving the processing to `DefaultTest`, which is renamed to `RoundOrchestrator`
- [x] Move the round-related objects (monitor, report, etc.) to `RoundOrchestrator`, making `CaliperEngine` more lightweight (only flow control).
- [x] Instead of the modules loading the configuration objects themselves from a file, they should receive the already loaded object. This will make the testing easier (no more fake YAML loader mocks).
- [x] Instead of passing around configuration __file paths__, the modules could query them from the `Config` module (including the paths passed to the workers).
- [x] Extract worker process message handling into a base class.


aklenik:
@nklincoln These are the changes I plan in this PR. I'll add commits gradually, so it's easier to follow the changes (I'll squash at the end). Plus it's easier to run every test through the CI, hence the draft PR :)

aklenik:
@nklincoln The PR is ready for review. Let me know when it's time to squash

PR #650, opened 2019-11-14
Enable charts to be included in output report

nklincoln:
This PR provides enhanced report outputs for a user

Breaking changes:
- monitor configuration for docker and process monitors.
  - docker monitor `name` filter changed to `container`, to better reflect what it is filtering
  - process monitor now has the array of process to monitor under a `processes` block

Other Changes:
- Main report styling change
- Conversion of report tables to be normalised to a single metric. Each column is normalised to the largest value, to a user configurable precision. 
- Addition of charts to output report (user opt-in, for each monitor)

### DIFF
--- a/packages/caliper-core/lib/master/charts/chart-builder.js
+++ b/packages/caliper-core/lib/master/charts/chart-builder.js
@@ -1,0 +1,252 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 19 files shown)


packages/caliper-core/lib/master/charts/chart-builder.js
```diff
@@ -0,0 +1,252 @@
...
+                        data: [obj0_data, obj1_data, obj2_data, ...],
+                    }
+                ]
+            })
+        }
+     *
+     * @param {string} callingMonitor the calling class name used to create a UUID
+     * @param {string} testLabel the current test label, required for creating a UUID
+     * @param {string[]} includeMetrics the metrics to include in the chart
+     * @param {Map<string, string>} resourceStats all resource stats
```
aklenik: Param type should be `Map<string, string>[]`

PR #652, opened 2019-11-14
Add Ethereum Support for HD Keys

shemnon:
This PR adds support for each Caliper client to have its own Ethereum
account to be deterministically derived from a seed value using BIP44 style
key derivation. This is only supported for the `fromAddress` value.  

This enables besu support of multiple caliper clients instead of single
client operations.

To use this instead of specifying a `fromAddresss` and an optional private
key instead a `fromAddressSeed` value is specified, which is a 128 bit hex
encoded value used as the seed for key derivation.  Each client uses its
client index as the account field in the BIP-44 path.  In the future where multiple
accounts are needed per caliper client the address_index field will be
incremented or mapped to the account value.

The genesis file in the samples and integration tests for besu have had
the first 10 keys pre-seeded with test ether.  The sample and
integration test for besu has also been updated to use a from address
seed.

Signed-off-by: Danno Ferrin <danno.ferrin@gmail.com>

## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/caliper)


shemnon:
So one config file per feature in the integration?  The integration test prior to the latest reboot had both explicit gas and estimated gas tests, I can set that up in separate tests, so I see 3 - estimated gas, explicit gas, and multi client HD wallets.

aklenik:
@shemnon Yeah, something like that. The Fabric tests use the same network for every phase (start network, perform different stuffs different ways, shutdown network). I don't know whether this is feasible for the Ethereum/Besu tests. Should be, even if you deploy the same contract multiple times on different addresses.

shemnon:
Integration tests updated (right now it's single client then dual client, we get the coverage we need).  Docs are updated in #666 and samples are updated in https://github.com/hyperledger/caliper-benchmarks/pull/33

PR #683, opened 2019-12-09
Document the messenger use

nklincoln:
Documentation for possible messenger implementations between the orchestrator and workers


nklincoln:
Just a comment to make stalebot go away

PR #696, opened 2020-01-06
Integration test for benchmark generator

Lynn14m:
Signed-off-by: lmuswere <lynn14m@gmail.com>


.travis/benchmark-integration-test-verdaccio.sh
```diff
@@ -22,6 +22,7 @@ npm i && npm run repoclean -- --yes && npm run bootstrap
 
 # Call CLI through the local binary
 export CALL_METHOD="npx caliper"
+export GENERATOR_METHOD="yo"
```
nklincoln: you will need the explicit generator call here ... ie `yo caliper-generator` etc

packages/caliper-tests-integration/generator_tests/myWorkspace/src/marbles/node/package.json
```diff
@@ -0,0 +1,17 @@
+{
+	"name": "marbles",
+	"version": "1.0.0",
+	"description": "marbles chaincode implemented in node.js",
+	"engines": {
+		"node": ">=8.4.0",
+		"npm": ">=5.3.0"
+	},
+	"scripts": {
+		"start": "node marbles.js"
```
nklincoln: general comment - is it necessary to include the smart contracts in the workspace? Is it possible to link direct to existing s-contracts outside the workspace, or is that a current limitation of caliper (if so we might want to fix that at some point ... but not for this)

nklincoln:
👌 Excellent ... the final piece of the generator puzzle 👍

PR #703, opened 2020-01-12
[Built on top of messenger PR] Split Fabric Adaptors and enable Fabric v2

### LINKED ISSUE #684: Account for breaking changes in Fabric v2.0
Fabric V2 has some big changes that will need to be accounted for, whilst maintaining backward compatibility so that it is still possible to benchmark V1.x networks.

The most appropriate way to enable this is to follow the same pattern as used in the modularisation of the RateControllers/observers etc. That would enable us to dynamically load the correct (versioned) adaptor based on the binding version that has been used.

nklincoln:
This PR relies on #682 

Splits the Fabric adaptors so that there are two code bases instead of a single file that deals with different possible transaction routes. There are now separate files for:
- standard caliper test path using node SDK
- test path using Fabric Gateway

Adaptors are further divided to enable testing on SDKs with breaking changes, with semver checks being made in order to select the correct version adaptor.

The CLI has **not** been modified to bind to a Fabric V2 SDK, since there are no stable published versions, however it has been tested on the most recent snapshot available on NPM

Other changes:
- Gateway adaptor has been enabled to select from inbuilt SDK event/query strategy using flags, which default to the SDK default options
- `getType()` has been added to the Blockchain interface

closes #684


packages/caliper-fabric/lib/fabric.js
```diff
@@ -14,2181 +14,57 @@
...
-                    }
-                }
-
-                // check private collection configuration
-                if (CaliperUtils.checkProperty(ccObject, 'collections-config')) {
-                    if (this.version.lessThan('1.2.0')) {
-                        throw new Error(`${chaincodeInfo.id}@${chaincodeInfo.version} in ${channel}: private collections are supported from Fabric v1.2`);
-                    }
-
-                    request['collections-config'] = ccObject['collections-con
```
Lynn14m: 'fabric-client' being removed, so will not work in future. Use 'fabric-ca-client' instead.

nklincoln: good spot 👍

nklincoln: latest push bases the conditioning on fabric-ca-client 👍

nklincoln:
wohoo!!

PR #734, opened 2020-02-18
Provide cli command to spawn a distributed worker 

### LINKED ISSUE #697: Extend Caliper CLI
The current CLI enables:
- binding of an adaptor to a specific SDK version
- running a benchmark

Possible extensions to the CLI include:

**bind and start a client**
With the enablement of client messaging it is possible to stand up a single client, prior to the core master process being initiated. This will be useful for establishing a client within a docker container that is to be used for testing.

nklincoln:
CLI command to spawn a worker that uses a non-process comms method

Changes include:
- enable passage of yaml file to override internal default bind options
- worker spawn command conditioned on:
  - must be remote
  - must not use process based comms
- worker spawn includes calling of existing bind command
- Conversion of Worker factory to be a static method
- Rename of factories and sut workers from "client"

closes #697 


packages/caliper-cli/caliper.js
```diff
@@ -16,13 +16,15 @@
 'use strict';
 
 process.env.SUPPRESS_NO_CONFIG_WARNING = true;
+const { CaliperUtils } = require('@hyperledger/caliper-core');
+const Logger = CaliperUtils.getLogger('@hyperledger/caliper-cli');
```
aklenik: I'd use a simpler name for the logger, something like `CLI`, so it resembles a module name, rather than a package name.

packages/caliper-cli/lib/bind/bind.js
```diff
@@ -15,6 +15,8 @@
 'use strict';
 
 const { CaliperUtils, ConfigUtil } = require('@hyperledger/caliper-core');
+
+const fs = require('fs-extra');
```
aklenik: The simple `fs` package also has an `existsSync` function, or is there a reason we don't like that?
https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_fs_existssync_path

packages/caliper-cli/lib/bind/bind.js
```diff
@@ -34,31 +36,46 @@ class Bind {
...
+                bindOptions = CaliperUtils.parseYaml(file);
+            }
+        } else {
+            bindOptions = defaultBindOpts;
+        }
 
         let sutList = Object.keys(bindOptions.sut);
         if (!sut) {
-            let msg = `SUT name is not specified. Available SUTs: ${sutList.join(' | ')}`;
+            let msg = `SUT name is not specified. Available SUTs: ${Object.keys(defaultBindOpts.sut).join(' | ')}`;
```
aklenik: Shouldn't we list the SUT names available in the set config file (not necessarily the default)?
So `${Object.keys(bindOptions.sut).join(' | ')}` or simply ``${sutList.join(' | ')}``
Same for the other error messages.

packages/caliper-cli/lib/bindCommand.js
```diff
@@ -35,11 +31,11 @@ module.exports.builder = function (yargs){
...
-        'caliper-bind-cwd'  : {describe: 'The working directory for performing the SDK install', type: 'string'},
-        'caliper-bind-args'  : {describe: 'Additional arguments to pass to "npm install". Use the "=" notation when setting this parameter', type: 'string'}
+        'caliper-bind-cwd'  : {describe: 'The working directory for performing the SDK install', type: 'string' },
+        'caliper-bind-args'  : {describe: 'Additional arguments to pass to "npm install". Use the "=" notation when setting this parameter', type: 'string' },
+        'caliper-bind-file'  : {describe: 'Yaml file to override default (supported) package versions when binding an SDK', type: 'string' }
     });
     yargs.usage('Usage:\n  caliper bind --caliper-bind-sut fabric --caliper-bind-sdk 1.4.1 --caliper-bind-cwd ./ --caliper-bind-args="-g"');
-
     // enforce the option after these options
     yargs.requiresArg(['caliper-bind-sut','caliper-bind-sdk','caliper-bind-args', 'caliper-bind-cwd']);
```
aklenik: Hmm, the `requiresArg` call has been removed from our `run` command, so the options can also be specified from other sources (env, config file, etc). Strange how this line didn't cause a problem so far....

packages/caliper-cli/lib/worker/lib/launch.js
```diff
@@ -0,0 +1,58 @@
...
+        let sdk = ConfigUtil.get(ConfigUtil.keys.Bind.Sdk, undefined);
+
+        // only valid command if distributed workers
+        if (!ConfigUtil.get(ConfigUtil.keys.Worker.Remote, false)) {
+            const msg = 'Configuration definition indicates that worker is not remote: worker launch is invalid in this mode';
+            throw new Error(msg);
+        }
+
+        // not valid for process communications
+        if (ConfigUtil.get(ConfigUtil.keys.Worker.Communication.Method, false).localeCompare('process') === 0) {
```
aklenik: Shouldn't `false` be `process` instead? It's a string setting IIRC.
Also, can we derive this information using some built-in way? Like suggested here: https://stackoverflow.com/questions/30763112/how-to-tell-if-child-node-js-process-was-from-fork-or-not

The message transport detection/configuration still a bit more complicated than it should be (from a lazy user perspective). We should discuss it more, maybe we get to an automagical solution.

packages/caliper-cli/lib/bind/bind.js
nklincoln: this was an attempt to try and protect us from someone passing in an incorrect file - if it is completely wrong, then we won't be able to list anything.

Which raises the question on the need for a validator here

packages/caliper-cli/lib/bindCommand.js
nklincoln: the above are the required items as per the bind configuration only - which iirc also complains if the items are not passed. But you are correct - this is essentially a run command, so we should map to that pattern and only throw/reject if there are no values at the point of use

packages/caliper-cli/lib/worker/lib/launch.js
nklincoln: sounds like a good conversation to have on the call 👍

packages/caliper-cli/lib/bind/bind.js
aklenik: Yeah, a validator is definitely needed. Then let's leave it as it is for now.

packages/caliper-cli/caliper.js
```diff
@@ -16,24 +16,28 @@
...
 const version = 'v' + require('./package.json').version;
 
 let results = yargs
     .commandDir('./lib')
     .help()
-    .example('caliper bind\ncaliper benchmark run\ncaliper zooclient start\ncaliper zooservice start  ')
+    .example('caliper bind\ncaliper benchmark run\n caliper worker launch')
     .demand(1)
     .wrap(null)
     .strict()
```
aklenik: This can be removed, or the `strict(false)` should be moved here

nklincoln: good spot!!

PR #738, opened 2020-02-19
Fix Travis badge in README and simplify package README handling

aklenik:

* The Travis badge now points to the .com site
* The package READMEs have been removed. The publishing script copies the root README into every published package directory. So only one README to maintain.


aklenik:
Already merged, but github left this open...

PR #758, opened 2020-03-05
Add script to check/modify package versions

nklincoln:
Bumping versions will be a requirement going forwards, and we should have a script to do this.

This PR provides a script to enable checking version consistency across all packages, and fixing the versions through all packages.

Version level (desired/enforced) is set within the lerna.json file
- we check all packages match this using the command `npm run check-version`
- we fix all packages to the lerna.json file using the command `npm run fix-version`

Provided the lerna.json file is kept up to date, this should help us keep all package versions in step.

Also includes:
- align packages (all included in lerna file)
- align versions in all pacakges
- bump to node10 in travis


aklenik:
@nklincoln Maybe we should move the version check nodejs scripts into the publishing CLI. I've done this in the following branch with a `fix` command. You could expand that into `version fix` and `version check` subcommands. Thoughts?

aklenik:
Covered by #805

Issue #776, opened 2020-03-24
Ethereum adapter needs better nonce management


benjamincburns:
## Context
Ethereum transactions are required to contain a nonce. The nonce of a transaction is zero-based, and must be equal to the number of transactions previously submitted (and accepted into the mempool) by the account signing the transaction. If a transaction is submitted with an incorrect nonce, it is rejected before entering the mempool. This includes cases when transactions are submitted in rapid succession, but out of order.

## Expected Behavior
- Caliper will submit transactions in the correct order without nonce errors.
- Other transaction errors will not trigger nonce errors.

## Actual Behavior
- Caliper sometimes submits transactions out of order. When this happens, all outstanding transactions fail due to incorrect nonces.
- When transaction submission errors occur that cause the transaction to not enter the mempool, all following transactions for a given account fail.

## Possible Fix

There are actually three issues here.

1. Somehow async requests being made to the node's RPC interface are being reordered, causing them to arrive out of order.
2. The ethereum adapter does not detect transaction submission failures and update the state of its internal nonce tracking accordingly.
3. Many transactions are generated at once for the same account, meaning that even if nonce tracking were updated accordingly on transaction submission failures, many in-flight transactions will fail.

## Steps to Reproduce
1. Run a benchmark with a large transaction count against an ethereum client - it's most noticeable when running against ganache-cli, but it also occurs when benchmarking Besu.

## Context
This bug is impacting users' ability to benchmark Besu and other compatible ethereum client implementations. 

## Your Environment

* Version used: v0.3.0 and current master
* Environment name and version (e.g. Chrome 39, node.js 5.4): various nodejs versions from v8.17.0 to v11.19.0.
* Operating System and version (desktop or mobile): Ubuntu 19.10

benjamincburns:
Regarding out of order transactions, so far these are the solutions I've come up with. I'm not overly pleased with them however, so I'd appreciate suggestions for better alternatives.

1. Use the `async-mutex` to lock a mutex when the nonce is assigned to the transaction parameters, and unlock it when the transaction hash is received.
2. Switch to using web3 only for transaction encoding, and use node's lower-level HTTP apis to submit and monitor the transaction, applying the same mutex strategy from the previous solution, but unlock when the request is written to the client.
3. Move the transaction submission logic to a native module.

[...]

PR #780, opened 2020-03-31
ethereum: prevent unnecessary additional requests

benjamincburns:
Web3 has recently added features that lazily fetches missing information
when signing and sending transactions. In our case, this was the
chainId, and the current gasPrice. Because transactions are submitted
asynchronously, these additional requests would sometimes cause
transactions to be submitted out of order. Even if this probelm were
mitigated, these additional requests added spurious latency spikes to
some transactions, negatively impacting the test results.

This partially addresses #776  by eliminating request reordering issues
for stable websocket connections. This change does not however address
transaction reordering for HTTP connections, or for websocket
connections that require reconnects. It also does not solve the
"christmas tree light" issue w.r.t. nonce errors.

### DIFF
--- a/packages/caliper-ethereum/lib/ethereum.js
+++ b/packages/caliper-ethereum/lib/ethereum.js
@@ -101,11 +101,21 @@ class Ethereum extends BlockchainInterface {
      */
     async getContext(name, args) {
         let context = {
+            chainId: 1,
             clientIndex: this.clientIndex,
+            gasPrice: 0,
             contracts: {},
             nonces: {},
             web3: this.web3
         };
+
+        context.gasPrice = this.ethereumConfig.gasPrice !== undefined
+            ? this.ethereumConfig.gasPrice
+            : await this.web3.eth.getGasPrice();
+
+        context.chainId = this.ethereumConfig.chainId !== undefined
+            ? this.ethereumConfig.chainId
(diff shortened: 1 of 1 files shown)


packages/caliper-ethereum/lib/ethereum.js
```diff
@@ -212,6 +222,21 @@ class Ethereum extends BlockchainInterface {
                 context.nonces[context.fromAddress] = nonce + 1;
                 params.nonce = nonce;
             }
+
+            if (context.gasPrice !== undefined) {
```
benjamincburns: In retrospect this block doesn't need to be an if statement, as we'd just be assigning `undefined` if it were eliminated. It's benign to leave it in, but I'm happy to remove it if it makes people happy. Same comment for line 233.

shemnon:
LGTM

benjamincburns:
I spent most of yesterday testing out this patch and it seems to fix most of the problems we've observed (again, when using a websocket connection). Prior to this change, if we set the transaction send rate too high, we'd see a bunch of errors coming back. Now it works as you'd expect - there aren't any errors, and things just slow down. Personally I wouldn't trust benchmarking results from Caliper versions prior to this change, as they're very likely under reporting on performance.

PR #782, opened 2020-04-01
Fix CI workloads and enhance error handling

aklenik:

* The `bcType` => `getType()` fix is applied to the Ethereum and Besu CI workload modules.
* The TX and unexpected errors are separated in the Ethereum adapter. The latter now halt and fails the benchmark.
* Worker-side errors are now reported with a stack trace in the logs.
* Sawtooth adapter bug fixed, referenced non-existing instance variable.

### DIFF
--- a/packages/caliper-sawtooth/lib/sawtooth.js
+++ b/packages/caliper-sawtooth/lib/sawtooth.js
@@ -36,7 +36,7 @@ const {
 const request = require('request-promise');
 const _ = require('lodash');
 
-let configPath;
+let config;
 let lastKnownBlockId=null;
 //let blockCommitSatus = new Map();
 let batchCommitStatus = new Map();
@@ -48,7 +48,6 @@ let currentEndpoint= 0;
 * @return {String} rest endpoint url
 */
 function getRESTUrl() {
-    let config = require(configPath);
     let restApiUrls = config.sawtooth.network.restapi.urls;
     currentEndpoint++;
     if(currentEndpoint >= restApiUrls.length) {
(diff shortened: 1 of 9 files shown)


packages/caliper-sawtooth/lib/sawtooth.js
```diff
@@ -365,7 +365,7 @@ class Sawtooth extends BlockchainInterface {
      * @return {Promise} The return promise.
      */
     getContext(name, args) {
-        let config  = require(this.configPath);
+        let config  = require(configPath);
```
nklincoln, line 368: This one kinda makes me itchy. I think the better solution here would be to set the config inside the Sawtooth constructor, and move the functions into the class so that they have access to a `this.config` property. Moreover, this would remove dynamic requires in the "hot path" during `invokeSmartContract`

aklenik:
@nklincoln I decided to make the config object a global variable, just like the configPath was. It was easier than refactoring the entire adapter 😅 
@shemnon @benjamincburns Please double check that I correctly merged your previous PR, because I made some structural changes in the TX send function.

benjamincburns:
> @nklincoln I decided to make the config object a global variable, just like the configPath was. It was easier than refactoring the entire adapter 😅
> @shemnon @benjamincburns Please double check that I correctly merged your previous PR, because I made some structural changes in the TX send function.

It looks like these changes incorporate my work from #780 correctly.

Does making the `config` object a module-scoped var in the sawtooth adapter hurt testability? I'm not too familiar with the code here, but I can imagine it impacting the ability to do concurrent unit tests, for example.

aklenik:
> Does making the `config` object a module-scoped var in the sawtooth adapter hurt testability? I'm not too familiar with the code here, but I can imagine it impacting the ability to do concurrent unit tests, for example.

@benjamincburns You are right, the structure of the module could be improved (by making it more OO). It should be a trivial refactoring, but currently, most of our efforts go into cleaning up and enhancing the core code, and not the adapters (yet). But feel free to refactor it if you have some extra free time :) Although, working on the Ethereum/Besu adapter is probably more useful based on the community activity.

nklincoln:
ok, I'll raise a separate issue to refactor the sawtooth adaptor 👍

shemnon:
Also, while in sawtooth code you missed a couple of `.bcType`->`.getType()` changes, one in query.js and one in smallbankoperation.js

PR #783, opened 2020-04-02
Enable fabric adaptor with most recent snapshot of v2 fabric sdk

nklincoln:

A few changes here:
- Bindings
  - Added a bind point with a current V2 snapshots
- ESLint
  - Added `prefer const` to the eslint rule for fabric
- Fabric V2
  - Will throw if not using `gateway` mode (can work on non-gateway in background and enable it)
  - Updated fabric gateway adaptor to be compatible with latest working snaphot
    - No admin actions (warn if used)
    - Account for breaking changes in APIs
  - Added a registrar helper to perform any enrol actions for users to with no credentials
  - Enabled `countQueryAsLoad`
- Fabric V1 Gateway
  - Updated the adaptor to mirror the V2 adaptor and accept transient data
  - Enabled `countQueryAsLoad`

### DIFF
--- a/packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
+++ b/packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
@@ -143,7 +143,7 @@ class Fabric extends BlockchainInterface {
 
         this.network = undefined;
         if (typeof networkConfig === 'string') {
-            let configPath = CaliperUtils.resolvePath(networkConfig, workspace_root);
+            const configPath = CaliperUtils.resolvePath(networkConfig, workspace_root);
             this.network = CaliperUtils.parseYaml(configPath);
         } else if (typeof networkConfig === 'object' && networkConfig !== null) {
             // clone the object to prevent modification by other objects
@@ -209,16 +209,16 @@ class Fabric extends BlockchainInterface {
      * @private
      */
     _assembleTargetEventSources(channel, targetPeers) {
-        let eventSources = [];
+        const eventSources = [];
         if (this.networkUtil.isInCompatibilityMode()) {
             // NOTE: for old event hubs we have a single connection to every peer set as an event source
             const EventHub = require('fabric-client/lib/EventHub.js');
 
-            for (let peer of targetPeers) {
-                let org = this.networkUtil.getOrganizationOfPeer(peer);
-                let admin = this.adminProfiles.get(org);
+            for (const peer of targetPeers) {
+                const org = this.networkUtil.getOrganizationOfPeer(peer);
+                const admin = this.adminProfiles.get(org);
 
-                let eventHub = new EventHub(admin);
+                const eventHub = new EventHub(admin);
                 eventHub.setPeerAddr(this.networkUtil.getPeerEventUrl(peer),
                     this.networkUtil.getGrpcOptionsOfPeer(peer));
 
@@ -229,11 +229,11 @@ class Fabric extends BlockchainInterface {
                 });
             }
         } else {
-            for (let peer of targetPeers) {
-                let org = this.networkUtil.getOrganizationOfPeer(peer);
-                let admin = this.adminProfiles.get(org);
+            for (const peer of targetPeers) {
+                const org = this.networkUtil.getOrganizationOfPeer(peer);
+                const admin = this.adminProfiles.get(org);
 
-                let eventHub = admin.getChannel(channel, true).newChannelEventHub(peer);
+                const eventHub = admin.getChannel(channel, true).newChannelEventHub(peer);
 
                 eventSources.push({
                     channel: [channel], // unused during chaincode instantiation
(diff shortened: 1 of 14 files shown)


packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
```diff
@@ -1497,73 +1481,69 @@ class Fabric extends BlockchainInterface {
...
-        invokeStatus.Set('request_type', 'query');
+        const invokeStatus = new TxStatus(uuid());
+
+        // Add transient data if present
+        // - passed as key value pairing such as {"hello":"world"}
+        if (invokeSettings.transientData) {
+            const transientData = {};
+            const keys = Array.from(Object.keys(invokeSettings.transientData));
+            keys.forEach((key) => {
+                transientData[key] = Buffer.from(transientData[key]);
```
aklenik, line 1508: I think it should be `invokeSettings.transientData[key]` when creating the buffer. Same for v2 gateway code, but I can't add a comment there :)

packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
```diff
@@ -1497,73 +1481,69 @@ class Fabric extends BlockchainInterface {
...
-        const client = this.clientProfiles.get(querySettings.invokerIdentity);
+        // Create a transaction
+        const transaction = smartContract.createTransaction(invokeSettings.chaincodeFunction);
 
         // Build the Caliper TxStatus, this is a reduced item when compared to the low level API capabilities
-        const txIdObject = client.newTransactionID();
-        const txId = txIdObject.getTransactionID();
-        let invokeStatus = new TxStatus(txId);
-        invokeStatus.Set('request_type', 'query');
+        const invokeStatus = new TxStatus(uuid());
```
aklenik, line 1500: Don't we have access to the TX ID somewhere? This way it's nearly impossible to match client-side and Fabric-side logs, which is a bummer for data/log analysis.

packages/caliper-fabric/lib/adaptor-versions/v2/fabric-gateway-v2.js
```diff
@@ -1692,41 +579,20 @@ class Fabric extends BlockchainInterface {
...
     async init(workerInit = false) {
-        let tlsInfo = this.networkUtil.isMutualTlsEnabled() ? 'mutual'
+        const tlsInfo = this.networkUtil.isMutualTlsEnabled() ? 'mutual'
             : (this.networkUtil.isTlsEnabled() ? 'server' : 'none');
-        let compMode = this.networkUtil.isInCompatibilityMode() ? '; Fabric v1.0 compatibility mode' : '';
-        logger.info(`Fabric SDK version: ${this.version.toString()}; TLS: ${tlsInfo}${compMode}`);
+        logger.info(`Fabric SDK version: ${this.version.toString()}; TLS: ${tlsInfo}`);
 
+        logger.warn(`Administrative actions are not possible with Fabric SDK version: ${this.version.toString()}`);
         await this._initAdaptor(workerInit);
```
aklenik, line 587: Passing `workerInit` is pointless, since it's missing from the signature of `_initAdaptor()`

packages/caliper-fabric/lib/adaptor-versions/v2/fabric-gateway-v2.js
```diff
@@ -1559,12 +445,7 @@ class Fabric extends BlockchainInterface {
         const transaction = smartContract.createTransaction(invokeSettings.chaincodeFunction);
 
         // Build the Caliper TxStatus, this is a reduced item when compared to the low level API capabilities
-        const txId = transaction.getTransactionID();
-        let invokeStatus = new TxStatus(txId);
-
-        if (context.engine) {
-            context.engine.submitCallback(1);
-        }
+        const invokeStatus = new TxStatus(uuid());
```
aklenik, line 448: Same TX ID question here.

packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
nklincoln: The SDK has changed now, in the the txID is created mid flow and no longer passed in. It should be possible to pull out the txID at the end and update the completed transaction ... which would yield a similar ability

aklenik: That's fine by me, as long as we have traceability. The Ethereum adapter also sets it at the end :+1:

packages/caliper-fabric/lib/adaptor-versions/v1/fabric-gateway-v1.js
nklincoln: good spot!

packages/caliper-fabric/lib/adaptor-versions/v2/fabric-gateway-v2.js
nklincoln: agree with that one, will remove

PR #786, opened 2020-04-03
Update Besu Support

shemnon:
Update besu to support 1.3 and 1.4 via semver.  Keep 1.3.2 in place to
support legacy configurations.

Signed-off-by: Danno Ferrin <danno.ferrin@gmail.com>

### DIFF
--- a/packages/caliper-cli/lib/lib/config.yaml
+++ b/packages/caliper-cli/lib/lib/config.yaml
@@ -96,8 +96,10 @@ sut:
         latest: *ethereum-latest
 
     besu:
-        1.3.2: &besu-latest
+        1.3.2:
             packages: ['web3@1.2.0']
-        1.3: *besu-latest
-        1.4: *besu-latest
+        1.3: 
+            packages: ['web3@1.2.0']
+        1.4: &besu-latest
+            packages: ['web3@1.2.0']
         latest: *besu-latest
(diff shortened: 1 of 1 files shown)


packages/caliper-cli/lib/lib/config.yaml
```diff
@@ -98,4 +98,6 @@ sut:
     besu:
         1.3.2: &besu-latest
             packages: ['web3@1.2.0']
+        1.3: *besu-latest
```
nklincoln, line 101: besu-latest should only be applied to 1.4 👍

shemnon: I'm not a YAML expert.  How should this look?  All of 1.3.2, 1.3, 1.4, and latest use the exact same dependencies.  & is a reference and * is an expansion of that reference.  Do you want a different name or do I not understand the syntax?

aklenik: Technically, the YAML is correct. Although, it is a bit confusing to read that the `besu-latest` anchor is used as an alias at multiple places. 

1. The quick&dirty solution is to list the package explicitly at both places, then put the `besu-latest` anchor to `1.4` and use the alias at `latest`. This will allow the independent modification of 1.3 and 1.4 dependencies (whether it'll be needed, that's another question). 
2. If you don't want to duplicate entries, place a single `packages: [...]` entry at the top of the file, under an attribute and anchor like `besu-web3-1.2.0`, and reference it for 1.3 and 1.4, while also attaching the `besu-latest` anchor to 1.4 (hopefully anchors and aliases at the same place are supported). Although it seems a bit overengineered for the current situation. 

My take on this is to go for readability (1). Listing the package at both places is not the end of the world, but it's easier to read.

PR #792, opened 2020-04-08
update v2 gateway

nklincoln:

Make use of the most recent ability to gain direct access to the txID from the transaction object.
- updated v2 binding
- removed uuid require
- txID is populated on the TxStatus object after the txn has been completed
- cried when I noticed that the TxStatus object uses capitalised method names 😢


packages/caliper-fabric/lib/adaptor-versions/v2/fabric-gateway-v2.js
```diff
@@ -489,6 +489,7 @@ class Fabric extends BlockchainInterface {
             invokeStatus.result = result;
             invokeStatus.verified = true;
             invokeStatus.SetStatusSuccess();
+            invokeStatus.SetID(transaction.getTransactionId());
```
aklenik, line 492: The TX ID should be set also for failed TXs.

nklincoln: 👍 Change made

PR #808, opened 2020-04-18
ethereum: add contractDeployerAddress to context

benjamincburns:
Without this change, it's impossible to configure Ownable contracts and similar from within the benchmark callback script. This is useful for RBAC contracts as well.


nklincoln:
@benjamincburns - thanks for the contribution 👍

PR #825, opened 2020-05-05
ethereum: revert erroneous change to web3 dependency for ethereum adapter

benjamincburns:

### DIFF
--- a/packages/caliper-ethereum/package.json
+++ b/packages/caliper-ethereum/package.json
@@ -26,7 +26,7 @@
         "ethereumjs-wallet": "^0.6.3"
     },
     "devDependencies": {
-        "web3": "0.20.7",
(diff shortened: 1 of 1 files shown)


packages/caliper-ethereum/package.json
```diff
@@ -26,7 +26,7 @@
         "ethereumjs-wallet": "^0.6.3"
     },
     "devDependencies": {
-        "web3": "0.20.7",
+        "web3": "1.2.2",
```
kiview, line 29: What about directly updating to 1.2.7?

benjamincburns: It's presently broken, so this change is just to revert to the way it was before it was broken. I'm not opposed to updating, but I don't think that should be prompted by this sort of breakage.

PR #829, opened 2020-05-06
ethereum: disallow http(s) RPC transports

### LINKED ISSUE #776: Ethereum adapter needs better nonce management

benjamincburns:
Fixes #776. See https://github.com/hyperledger/caliper/issues/776#issuecomment-624771622 for more info.

### DIFF
--- a/packages/caliper-ethereum/lib/ethereum.js
+++ b/packages/caliper-ethereum/lib/ethereum.js
@@ -38,12 +38,41 @@ class Ethereum extends BlockchainInterface {
      */
     constructor(workerIndex) {
         super();
-        let configPath = CaliperUtils.resolvePath(ConfigUtil.get(ConfigUtil.keys.NetworkConfig));
         this.bcType = 'ethereum';
-        this.ethereumConfig = require(configPath).ethereum;
+
+        let configPath = CaliperUtils.resolvePath(ConfigUtil.get(ConfigUtil.keys.NetworkConfig));
+        let ethereumConfig = require(configPath).ethereum;
+
+        // throws on configuration error
+        this.checkConfig(ethereumConfig);
+
+        this.ethereumConfig = ethereumConfig;
         this.web3 = new Web3(this.ethereumConfig.url);
         this.web3.transactionConfirmationBlocks = this.ethereumConfig.transactionConfirmationBlocks;
         this.clientIndex = workerIndex;
+    }
+
+    /**
+     * Check the ethereum networkconfig file for errors, throw if invalid
+     * @param {object} ethereumConfig The ethereum networkconfig to check.
+     */
+    checkConfig(ethereumConfig) {
+        if (!ethereumConfig.url) {
+            throw new Error(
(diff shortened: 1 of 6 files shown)


benjamincburns:
The CI errors appear to be caused by something to do with the prometheus observer.

Unless there's some code in those tests that uses the ethereum adapter via an HTTP provider, it's impossible that this change could have caused this error. I'll double check for usages of `caliper-ethereum` in the failing tests just to be sure.

benjamincburns:
My apologies - I missed the linter errors - should be fixed up now.

nklincoln:
There's a connection error showing at the moment - is this a port mismatch? The config files (docker compose) used in the integration test are exposing/mapping a single port (8545) rather than a port range as done in besu, and there has been a change in the network config files to be 8546

(https://github.com/hyperledger/caliper/blob/9b571dd2f48ef745ff929c1cb1daebeaa998dae7/packages/caliper-tests-integration/ethereum_tests/config/docker-compose.yml#L26)

benjamincburns:
> There's a connection error showing at the moment - is this a port mismatch? The config files (docker compose) used in the integration test are exposing/mapping a single port (8545) rather than a port range as done in besu, and there has been a change in the network config files to be 8546
> 
> (
> 
> https://github.com/hyperledger/caliper/blob/9b571dd2f48ef745ff929c1cb1daebeaa998dae7/packages/caliper-tests-integration/ethereum_tests/config/docker-compose.yml#L26
> 
> )

Good catch - I forgot to change the exposed port.

PR #830, opened 2020-05-06
ethereum config: document websocket and gas limit changes

benjamincburns:
Also includes some very small grammar and changes to the order of existing paragraphs to improve flow/readability.


docs/vNext/Ethereum_Configuration.md
```diff
@@ -59,7 +64,7 @@ The top-level `caliper` attribute specifies the type of the blockchain platform,
 Furthermore, it also contains two optional commands: a `start` command to execute once before the tests and an `end` command to execute once after the tests. Using these commands is an easy way, for example, to automatically start and stop a test network. When connecting to an already deployed network, you can omit these commands.
 
 These are the keys to provide inside the configuration file under the `ethereum` one:
-* [URL](#url) of the node to connect to. Only http is currently supported.
+* [URL](#url) of the node to connect to. Only websockets are currently supported.
```
kiview, line 67:
```suggestion
* [URL](#url) of the node to connect to. Only websocket is currently supported.
```
I would personally use the singular here (also below).

benjamincburns: Thanks - I'll update (I don't think I can use the "Commit suggestion" button because of the DCO requirement).

Issue #832, opened 2020-05-08
caliper benchmark generator cannot be installed via npm


mxevgenis:
I have followed the guide in https://hyperledger.github.io/caliper/vNext/benchmark-generator/ and in the second step where i try to install the generator-caliper i face the issue below:

0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node',
1 verbose cli   '/usr/bin/npm',
1 verbose cli   'install',
1 verbose cli   '-g',
1 verbose cli   '@hyperledger/generator-caliper' ]
2 info using npm@6.14.4
3 info using node@v10.20.1
4 verbose npm-session dfbe2f77ade8a42c
5 silly install loadCurrentTree
6 silly install readGlobalPackageData
7 http fetch GET 404 https://registry.npmjs.org/@hyperledger%2fgenerator-caliper 1030ms
8 http fetch GET 404 https://registry.npmjs.org/@hyperledger%2fgenerator-caliper 819ms
9 silly fetchPackageMetaData error for @hyperledger/generator-caliper@latest 404 Not Found - GET https://registry.npmjs.org/@hyperledger%2fgenerator-caliper - Not found
10 timing stage:rollbackFailedOptional Completed in 2ms
11 timing stage:runTopLevelLifecycles Completed in 1872ms
12 verbose stack Error: 404 Not Found - GET https://registry.npmjs.org/@hyperledger%2fgenerator-caliper - Not found
12 verbose stack     at res.buffer.catch.then.body (/usr/lib/node_modules/npm/node_modules/npm-registry-fetch/check-response.js:104:15)
12 verbose stack     at process._tickCallback (internal/process/next_tick.js:68:7)
13 verbose statusCode 404
14 verbose pkgid @hyperledger/generator-caliper@latest
15 verbose cwd /home/user/caliper-benchmarks
16 verbose Linux 4.4.0-116-generic
17 verbose argv "/usr/bin/node" "/usr/bin/npm" "install" "-g" "@hyperledger/generator-caliper"
18 verbose node v10.20.1
19 verbose npm  v6.14.4
20 error code E404
21 error 404 Not Found - GET https://registry.npmjs.org/@hyperledger%2fgenerator-caliper - Not found
22 error 404
23 error 404 '@hyperledger/generator-caliper@latest' is not in the npm registry.
24 error 404 You should bug the author to publish it (or use the name yourself!)
25 error 404 Note that you can also install from a
26 error 404 tarball, folder, http url, or git url.
[...]

PR #843, opened 2020-05-18
Publish Caliper benchmark generator

Lynn14m:
In relation to issue #832 

We want to have a published generator package called @hyperledger/generator-caliper. For Yeoman to be able to find the generator using the file system, the generator code must be in a folder with a name beginning with `generator-`. Initially the `generator-caliper` folder was located inside a `caliper-generator` folder to fit with the current repository structure. This structure needed to be changed in this PR, however, in order to be able to publish the generator without overly complicating the publish scripts.


nklincoln:
Shame that we cannot easily retain the caliper-x folder structure, but seems that yoeman scrapes repositories based on name and then checks the package tags for key-words. Let's hope that their undocumented "we find scoped" packages is working 😂

PR #852, opened 2020-06-03
Add semver check on 1.x sdk path

### LINKED ISSUE #847: Caliper will permit the use of gateway with SDKs that do not contain code
When using the `caliper launch master` command with the `caliper-fabric-gateway-usegateway` option and an SDK version without gateway code, we expect the command to not run and fail straight away as the code path doesn't exist in the SDK.

Instead the command is still able to be run and errors with `Unexpected error during benchmark execution: TypeError: Cannot read property 'MSPID_SCOPE_SINGLE' of undefined`

nklincoln:
closes #847 

Add a `>=1.4.2` semver check on the 1.x Fabric SDK pathway


nklincoln:
Invalidated by #863

PR #856, opened 2020-06-04
0.4.0 PR - Convert workload modules to classes

### LINKED ISSUE #576: Convert Test Callbacks To Classes
The current client callbacks, whilst functional, are not the best approach. They currently exist as simple javascript files, with exports. It would be far better if each callback was actually a class file, that implements an interface.

We could then initialise the class with a blockchain, context and arguments, which would leave the `init`, `run`, and `end` items to be implemented by the user.

This would remove the necessity of a user to explicitly save off important information as standard in the init function as it would be default be taken care of. 

If we consider the 'super' that would be called, we would ideally be setting the following information:
- context
- blockchain object on which to invoke/query
- target chaincodeID
- target chaincode function

The latter two are currently hard coded in the callback, and can be much more elegantly set within the benchmark configuration file, ie:

```
- label: myLabel
    description: This is my description, there are many like it, but this one is mine
    chaincodeId: targetChaincode
    chaincodeFunc: targetFunction
    txDuration: 300
    rateControl: { type: fixed-rate, opts: { tps: 15} }
    arguments:
      myArg: 8000
    callback: myCallbackObject
```

Might also want to consider changing the name of `callback` perhaps. But is we do the above, we remove most of the boiler plate code a user is forced into creating. The user will only have to tailor it for population of the transaction body.

Note we will need to modify the generator too - but in this case it is simplifying it.

Considering the above, the base class users could extend would include:
- constructor, which sets [context, blockchain object, chaincodeId, chaincodeFunc]
- empty init() function
- run that barfs if not overridden
- empty end() function
- invoke method
- query method

So a user would be able to:
- call super to set internal required items
- perform their own `init` if required
- override the run() method, build a txn body (if needed) and use `this.invoke()` or `this.query`
- perform their own `end` if required

aklenik:
- Create interface and base class for workload modules
- Load workload modules with the factory approach
- Update CI test workloads to class format
- Extend CI workload arguments where needed to make workloads deterministic
- Increase frequency of TX updates in CI tests to increase monitor accuracy
- Update workload module paths in the benchconfig files to precise relative path
- Create separate artifacts for Fabric v2 CI phase

Closes #576

Note: the generator integration test is broken currently, but maybe we should wait for the next workload module updates before adjusting it.

Docs update and benchmark repo update WIP.


aklenik:
@benjamincburns I've updated the ethereum and besu bindings to use web3@1.2.2, since 1.2.0 doesn't seem to contain the getChainId function (#838).
I've also updated the `packages/caliper-tests-integration/besu_tests/config/docker-compose.yml` file to enable ws connections. 

Could you please take a look at the ethereum-related changes? There still seems to be some nonce/gas? related errors for the ethereum CI test (besu runs fine): https://travis-ci.com/github/hyperledger/caliper/jobs/343851469#L661

Issue #870, opened 2020-06-15
MaxTPS rate controller


nklincoln:
There are already quite a few rate controllers, however a missing controller is that which seeks to find the maximum possible TPS that may be driven.

We currently have rate limiting controllers that seek to maintain a stead backlog of transactions such that the SUT is not overwhelmed by the transactions, however it is not necessarily going to tune itself to output the maximum achievable TPS; just the TPS that satisfies the backlog.

During the rate control cycle, we have access to the current TPS as witnessed by the worker. Meaning that is would be possible to step up (ramp) the TPS until the achieved rate drops below the send rate, backoff again and increase with a smaller ramp gradient.

This could in theory enable the maximum TPS to be investigated, since once all sending clients achieve an equilibrium, the workers are at their maximum driving rate.

aklenik:
Yes, a max-rate controller would be useful. However, finding a meaningful metric for the step up/down decisions will be hard.

For example, even if the txs still finish within a given latency threshold (one possible metric), the SUT could still be in an overloaded state (whatever that means). So I'm not sure that the limited information of the workers will be enough. But we can experiment with basic metrics.

(Btw, an unconditional max-rate/no-sleep controller is a perfect combination with a dummy test adapter to profile/stresstest the worker processes)

nklincoln:
closed by above PRs

PR #874, opened 2020-06-23
Add maximum-rate controller

nklincoln:

Contributes to #870 
- Ramp the worker TPS conditionally between txUpdate phases, if a TPS drop (for an increased driving rate) is detected, back off and reduce the ramp

### DIFF
--- a/packages/caliper-core/lib/worker/rate-control/maxRate.js
+++ b/packages/caliper-core/lib/worker/rate-control/maxRate.js
@@ -1,0 +1,221 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const RateInterface = require('./rateInterface.js');
+const Sleep = require('../../common/utils/caliper-utils').sleep;
+const Logger = require('../../common/utils/caliper-utils').getLogger('maxRate.js');
+
+/**
+ * Rate controller for driving at a maximum TPS.
+ */
+class MaxRate extends RateInterface {
+    /**
+     * Creates a new instance of the MaxRate class.
+     * @constructor
+     * @param {object} opts Options for the rate controller.
+     */
+    constructor(opts) {
+        super(opts);
+
+        // Default sleep
+        this.sleepTime = 100;
+
+        // Map for TPS observations
+        this.observedTPS = {
+            previous: 0,
+            current: 0
+        };
+
+        // Map for TPS settings
+        this.tpsSettings = {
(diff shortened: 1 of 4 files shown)


packages/caliper-core/lib/worker/rate-control/maxRate.js
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /**
+     * TPS from the previous txUpdate interval statistics
+     * @param {object} resultStats the passed stats object
+     * @return {number} the TPS within the interval
+     */
+    retrieveIntervalTPS(resultStats) {
+        const resultStatistics = resultStats[1];
+        return (resultStatistics.succ + resultStatistics.fail) / (resultStatistics.final.last - resultStatistics.create.min);
```
aklenik, line 173: Shouldn't we only consider the goodput? This way timed out TXs can also contribute to the overall TPS, while they normally indicate an overloaded system. 
Although ignoring expectedly failed TXs could speed up the send rate unintentionally... Insights?

packages/caliper-core/lib/worker/rate-control/maxRate.js
```diff
@@ -0,0 +1,203 @@
...
+     * @param {number} msg.clients The number of clients executing the round.
+     * @param {object} msg.clientArgs Arguments for the client.
+     * @param {number} msg.clientIdx The 0-based index of the current client.
+     * @param {number} msg.roundIdx The 1-based index of the current round.
+     *
+     * @async
+     */
+    async init(msg) {
+        this.sleepTime = 100;
+        this.reset = false;
```
aklenik, line 63: What's the purpose of `reset`?

packages/caliper-core/lib/worker/rate-control/maxRate.js
```diff
@@ -0,0 +1,203 @@
...
+     */
+    constructor(opts) {
+        super(opts);
+
+        // Map for TPS observations
+        this.observedTPS = new Map();
+        // Map for TPS settings
+        this.tpsSettings = new Map();
+        // MPS for observed stats
+        this.statisticsMap = new Map();
```
aklenik, line 38: Since the maps only contain a few specific keys, it might be more readable and efficient to use fix objects instead (with sensible default values):
```js
this.observedTPS = {
  previous: 0,
  current: 0,
  max: 0
}
```

nklincoln: I like!

packages/caliper-core/lib/worker/rate-control/maxRate.js
nklincoln: oops - that was left over from the prototype .. my bad!

packages/caliper-core/lib/worker/rate-control/maxRate.js
nklincoln: hmmm, not sure on this one. My feeling is that if we only account for goodput, we would not be accounting for all the transactions hitting the SUT .. and it might be the case that someone is intentionally wanting to investigate TPS rates with a certain percentage of failed transactions in the mix.

Maybe we need to enable clarification for the intended use of the "count as load" options when related to rate limiting and TPS determination?

aklenik: We could add a rate controller option like `countFailedTxs: true`. And then the user can choose the appropriate setting for the workload.

PR #875, opened 2020-06-23
Docs for max rate controller

### LINKED ISSUE #870: MaxTPS rate controller

nklincoln:

companion PR for #874 and closes #870

Docs for max rate controller use

### DIFF
--- a/docs/vNext/Rate_Controllers.md
+++ b/docs/vNext/Rate_Controllers.md
@@ -11,8 +11,9 @@ The rate at which transactions are input to the blockchain system is a key facto
 * [Fixed rate](#fixed-rate)
 * [Fixed feedback rate](#fixed-feedback-rate)
 * [Fixed backlog](#fixed-backlog)
-* [Composite rate](#composite-rate)
+* [Maximum rate](#maximum-rate)
 * [Linear rate](#linear-rate)
+* [Composite rate](#composite-rate)
 * [Zero rate](#zero-rate)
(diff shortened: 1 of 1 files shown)


docs/vNext/Rate_Controllers.md
```diff
@@ -59,6 +60,26 @@ The PID rate controller, targeting a backlog of 5 transactions with a starting T
 }
 ```
 
+## Maximum Rate
+The maximum rate controller is a controller for driving the workers to their maximum achievable rate. This controller will aim to maximize the driven TPS for the worker by ramping up the driven TPS and backing off again when a drop in TPS is witnessed.
```
aklenik, line 64: "maximum achievable rate without overloading the SUT" perhaps? Just to clarify that this is a smart controller :)

Issue #888, opened 2020-06-29
Consolidate connector API


aklenik:
## Context
As a prerequisite for adding TX builder support (#858), the connectors' TX API should be simplified/streamlined.

## Actual Behavior
1. Passing the context around between the workload module and the connector is meaningless. The workload gets the context during its initialization, the connectors also save the same instance. No need to pass it around.
2. The invoke/query APIs support batch requests, but the target contract, version, and timeout must be the same, only the arguments can differ. This is not realistic/flexible.
3. The invoke and query API is separated, but the difference is mainly on the SUT side, the client-side semantics/parameters are similar. Currently, it's difficult to mix invokes and queries in the same batch, since it requires the merging of result arrays in the workload modules.

## Expected Behavior
1. Remove the context from the API.
2. With the context removed, merge the contract, version, and timeout settings into the arguments object. This will result in a single `invokeSettings` parameter for the functions, which can also be an array.
3. The invoke and query function can be merged if we add a `query: true/false` attribute to the settings object above. This results in a single API function, expecting a single settings parameter, which can also be an array, now containing the mix of queries and invokes.

## Tasks
- [x] Remove context from API (#899)
- [x] Make TX reporting even oriented (#927)
- [x] Remove legacy `queryState` functions (#931)
- [x] Merge invoke/query parameters into a single object (#949)
- [ ] Merge invoke/query functions into a single function

## Your Environment

* Version used: v0.3.2

nklincoln:
At the moment the `init()` function is being used to initialise both the connector and the SUT for some adaptors. It would be good to remove this overloading by either having the connector init only on creation, or provide explicit `initConnector` and `initSUT` methods. My preference would be the latter due to some adaptors requiring an asynchronous init as they query for information on the SUT (Fabric)

aklenik:
I also support the `initConnector` and `initSUT` approach. I guess this means that worker processes won't even call the `initSUT` function. 

Separating the local and remote inits might not be that easy. For example, loading user connection profiles for Fabric sometimes requires enrolling new users, which requires SUT access. So separating these will need extra considerations and code. To this end, I'd open a separate issue for the init splitting. 

(The removal of `queryState` fits nicely into the invoke API refactor.)

aklenik:
Additional refactors in the connector API:
1. `get/releaseContext` is basically a `before/afterRound` call. If the context stays inside the connector, we should also hide the name.
[...]

PR #899, opened 2020-07-02
Remove the context parameter from the connector invoke/query API

aklenik:
Contributes to #888 


packages/caliper-iroha/lib/iroha-connector.js
```diff
@@ -298,10 +291,10 @@ class IrohaConnector extends BlockchainConnector {
             ) {
                 throw new Error('Invalid Iroha::getContext arguments');
             }
-            let context = {};
+            this.context = {};
```
nklincoln, line 294: seems unnecessary? Is instantly assigned in the next line

aklenik: True, that's an unintentional remnant 😅 Fixed.

PR #902, opened 2020-07-03
update git ignore file and replace report template

nklincoln:
Refactor of master -> manager neglected to update the gitignore file, which prevented the report template being found

This PR updates the gitignore file, and replace the report.html file


nklincoln:
fixed in #901

Issue #898, opened 2020-07-01
Multi channel support in fabric connector


nklincoln:
It would be nice to provide a more suitable structure to enable multi channel transactions within the Fabric connector. At the moment, whilst "kind of possible" it is not a route that is easily facilitated by Caliper.

## Possible Solution

There are a few changes required here. They will impact the Caliper configuration files and workload module (arguments), so now is the idea time to do this as we are making breaking changes

### Network Configuration Changes
1. Move the `wallet` used by (currently all) clients into the client definition. ie each client is linked to a wallet, meaning that clients may use the same *or different* wallets, thereby providing them with different identities. This is essential for organisational scoped clients, and useful for within organisations to provide explicit identities.
2. Within the channel definition, under each channel name, list the clients that have access to the channel. 

Through the above changes a user can:
- Provide clients with different wallets, which mirrors the current ability to provide clients with their own crypto material
- Explicitly name the channels that the client may interact with, without the need for discovery being enabled

### Fabric Connector Changes
Since it is possible to deploy the same name and version contract on different channels, we will need to build up a different hierarchy such that  *a client maps to multiple channels, these channels map to multiple contracts*. This implies the need to generate the hierarchy, which may be done in two ways:
- If using discovery, it is possible to determine the channel access for the client
- If not using discovery, it is possible use the channel/client list 

It is now required to be able to prescribe the channel to use for a transaction, which may be done through provision of a new transaction argument, such as `targetChannel`. If it is not provided, then it is assumed there is only one channel in use, and this default channel is used. 

Through the above changes a user can:
- Build a transaction to be submitted/evaluated on a named, or default, channel

aklenik:
Some questions/remarks regarding the issue.

## Wallets
I absolutely support the client-level wallet attribute. Flexibility++

## Channel access
Channel access in Fabric is determined on the organization level, right? We already know the MspId/Org of a client identity, so based on that we have an access list for channels. 

I think this is already used when initializing the channel objects of the CCPs. So I think we already have that information with the current config structure.

## Multiple channel support
Fabric contracts in the networkconfig have a globally unique `contractID` which defaults to the simple chaincode `id` when omitted. The adapter uses this to map a simple name to a channel-chaincode-version combination. Then only this simple name is referenced in the workload modules, so the channel mechanism of Fabric can be transparent.

Network config example from the CI test:
[...]

PR #916, opened 2020-07-08
move to organization based wallets for fabric connector

### LINKED ISSUE #882: Caliper gateway should report if an identity is missing from the wallet
If you are using the fabric adaptor, and nominate a wallet, the connector does not check for existence (certainly not within the v1 gateway) when initialising users. This can lead to very difficult to trace errors.

nklincoln:
Contributes to #898 

Closes #882 

Specified clients can now pick from `organizationWallets` placed in the high level network configuration file

Fabric validator modified to check `organizationWallets` and condition subsequent entries
Fabric connectors now contains a map of organisation wallets, within which organisation identities are stored/used

Integration test runs 'gateway' mode for V1 and V2 SDK, supplementary manual test has been run using minimal network configuration and file based wallets targeting VSCode based network, for both V1 and V2 fabric networks/SDK.

### DIFF
--- a/packages/caliper-fabric/lib/configValidator.js
+++ b/packages/caliper-fabric/lib/configValidator.js
@@ -14,6 +14,9 @@
 
 'use strict';
 
+const CaliperUtils = require('@hyperledger/caliper-core').CaliperUtils;
+const Logger = CaliperUtils.getLogger('config-validator');
+
 const j = require('@hapi/joi');
 
 /**
@@ -28,6 +31,7 @@ class ConfigValidator {
      * @param {boolean} gateway Indicates whether gateway mode is configured or not.
      */
     static validateNetwork(config, flowOptions, discovery, gateway) {
+        Logger.debug('Entering validateNetwork');
 
         // Current limitation is that default Caliper transactions do not work with discovery, since full network knowledge is required; it is required to use a gateway
         if (discovery && !gateway) {
@@ -51,7 +55,7 @@ class ConfigValidator {
         // see the lines: "tls = ... || nodeUrl.startsWith(...);"
 
         // can't validate mutual TLS now
-        ConfigValidator._validateTopLevel(config, flowOptions, discovery, gateway, tls);
+        ConfigValidator._validateTopLevel(config, flowOptions, discovery, tls);
 
         // validate CA section
         let cas = [];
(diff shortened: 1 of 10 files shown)


packages/caliper-fabric/lib/configValidator.js
```diff
@@ -513,17 +541,40 @@ class ConfigValidator {
...
+    }
+
+    /**
+     * Validates the given organization wallet configuration object.
+     * @param {object} config The configuration object.
+     */
+    static validateOrganizationWallet(config) {
+        Logger.debug('Entering validateOrganizationWallet');
+        const schema = j.object().keys({
+            path: j.string().min(1).optional(),
```
aklenik, line 554: If we have an org entry, then path should be required. Otherwise, the following is also valid:
```yaml
organizationWallets:
  org1:
  org2:
    path: 'path'
```

packages/caliper-fabric/lib/configValidator.js
```diff
@@ -112,12 +116,25 @@ class ConfigValidator {
             }
         }
 
+        // validate organizationWallets section
+        if (config.organizationWallets) {
+            const organizations = Object.keys(config.organizationWallets);
+            for (const organization of organizations) {
```
aklenik, line 122: The existence of an org should be checked. The previous code block gathers the valid org names, so it's a simple check.

(It might be better to validate the `organizationWallets` section as a whole, so rules like "valid keys", and "at least one key" can be specified. But it's not a dealbreaker, the above check is what matters.)

packages/caliper-fabric/lib/fabricNetwork.js
```diff
@@ -79,16 +75,16 @@ class FabricNetwork {
                 const cObj = this.network.clients[client].client;
                 // normalize paths
                 if (cObj.credentialStore) {
-                    cObj.credentialStore.path = CaliperUtils.resolvePath(cObj.credentialStore.path, workspaceRoot);
-                    cObj.credentialStore.cryptoStore.path = CaliperUtils.resolvePath(cObj.credentialStore.cryptoStore.path, workspaceRoot);
+                    cObj.credentialStore.path = CaliperUtils.resolvePath(cObj.credentialStore.path, this.workspaceRoot);
```
aklenik, line 78: Actually, `resolvePath` uses the configured workspace as the default value for a while now, so the `this.workspaceRoot` arguments are unnecessary. Since it's only used in `resolvePath`, it can be removed from the class altogether.
(I know it wasn't introduced by this PR, but it's a good place as any to clear it up 😄 )

packages/caliper-fabric/lib/connector-versions/v1/fabric-gateway.js
```diff
@@ -611,8 +609,8 @@ class Fabric extends BlockchainConnector {
         for (const org of orgs) {
 
             // providing registrar information is optional and only needed for user registration and enrollment
-            if (this.fileWalletPath) {
-                logger.info('skipping registrar initialization due to presence of file system wallet');
+            if (this.networkUtil.usesOrganizationWallets) {
```
aklenik, line 612: Function brackets `()` missing.

packages/caliper-fabric/lib/connector-versions/v2/registrarHelper.js
```diff
@@ -105,12 +112,15 @@ class RegistrarHelper {
...
      */
-    async registerUserForOrg(orgName, clientName) {
+    async registerUserForOrg(orgName, userID) {
+        logger.debug(`Entering registerUserForOrg for organization ${orgName} userID ${userID}`);
         const registrarInfo = this.registrarInfo.get(orgName);
-        const affiliation = this.networkUtil.getAffiliationOfUser(clientName);
-        return await this.registerUser(registrarInfo, clientName, { affiliation });
+        const affiliation = this.networkUtil.getAffiliationOfUser(userID);
+        const userSecret =  await this.registerUser(registrarInfo, userID, { affiliation });
+        logger.debug('Entering registerUserForOrg');
```
aklenik, line 122: Entering -> Exiting

PR #924, opened 2020-07-14
Refactor message handling

aklenik:
- Make the messenger interface event oriented
- Load messengers with factory methods
- Consolidate messenger implementations
- Create dedicated classes for every message type
- Make worker message handler object oriented
- Update messenger-adjacent modules to use the message classes and event callbacks
- Create constants for commonly used string identifiers
- Add explicit hyperledger/fabric-ccenv pulling and tagging
- Perform worker connection management only once, not for each round


packages/caliper-tests-integration/fabric_tests/caliper.yaml
```diff
@@ -42,7 +42,7 @@ caliper:
                 target: console
                 enabled: true
                 options:
-                    level: info
+                    level: debug
```
nklincoln, line 45: can result in travis log size being exceeded

force-clean.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
nklincoln, line 1: useful file but not called or exposed via npm script - did you intend to include it?

packages/caliper-tests-integration/fabric_tests/caliper.yaml
aklenik: You're right, I was just checking the usefulness of debug messages, will change it back

force-clean.sh
aklenik: Wasn't intentional, but I could wrap it in an npm script, since it's useful after a fabric rebinding. So revert or npm script should it be?

nklincoln: +1 for script

packages/caliper-tests-integration/fabric_tests/caliper.yaml
aklenik: Changed back.

force-clean.sh
aklenik: Added as npm script

Issue #919, opened 2020-07-08
Enable transaction.setEndorsingOrgs() in submit route for fabric connector


nklincoln:
It is possible to set an endorsing organization on a Fabric transaction. This can act as an alternative to setting endorsing peers, and (if using a gateway based transaction) will have the following result:
- when using discovery, will send to the listed peers, and handle failed transaction submission
- when not using discovery, will send to all peers

We already provide for setting `targetPeers` within the Fabric connector, it would be nice to enable this as an alternative

davidkel:
Completed now

PR #937, opened 2020-07-27
Add Support for targetOrganizations for fabric connector

davidkel:
Contributes to #919


packages/caliper-cli/lib/lib/config.yaml
```diff
@@ -60,10 +60,14 @@ sut:
             packages: ['fabric-client@1.4.6', 'fabric-protos@2.0.0-snapshot.1', 'fabric-network@1.4.6']
         1.4.7:
             packages: ['fabric-client@1.4.7', 'fabric-protos@2.0.0-snapshot.1', 'fabric-network@1.4.7']
-        1.4.8: &fabric-latest
+        1.4.8:
             packages: ['fabric-client@1.4.8', 'fabric-protos@2.0.0-snapshot.1', 'fabric-network@1.4.8']
-        2.1.0: &fabric-latest-v2
+        1.4.11: &fabric-latest
```
nklincoln, line 65: might as well add the 9/10 patch levels here. I think long term we need to consider if we only support latest patch levels for minor versions (only bind to 1.4.latest, 2.1.latest, etc)

Issue #928, opened 2020-07-17
As a Caliper user, I would like some tutorials to follow, so that I can understand how to use it


nklincoln:
## Context
Caliper has some documentation on components, but this is all in the form of a reference; it is largely left to users to follow/view integration tests and infer how to tailor it to their own needs. 

Ideally the Caliper documentation would include some tutorials on use, which would enable a user to follow in order to benchmark a selected SUT.

## Possible Fix
Starting in the vNext docs, we need a new LH section titled "Tutorials". That's the easy bit, the tutorial content needs to be decided.

Options:

1. Scenario based. We consider scenarios that people might be interested in, for instance:
    
    -  Benchmarking an existing Fabric network (picks up from end of BYFN, maybe also provides an ansible playbook to get there quickly)
    - Using Dockerized Caliper workers
    - Using Caliper in Kubernetes

2. Topic based

    - Writing a benchmark file
    - Writing a workload module
    - Using the <type> adapter

winslet:
An idea to add to the above:

Wonder if there's a quick "testing" example we could add to the above tutorials.

It could contain:
- a really lightweight ansible deployment of a network
- an example smart contract file
- an example caliper config
- anything else needed to get a testing example up and running

This gives people a jumping off point to modify for their needs as well as a "known good" example which could help them troubleshoot any problems theyre having with their own custom runs.

Thoughts?

aklenik:
@winslet Two approaches will clash here:
1. A user will start (for example) the BYFN tutorial in the `fabric-samples` repo, generating every artifact and starting the network. Then they would like to continue from there by adding Caliper-related artifacts and actually running Caliper. Jumping between two non-trivial samples/tools/repos can get messy, so they'll definitely need help there.
2. I absolutely support your idea that a self-contained `caliper-benchmarks`-only example could provide a clean starting point to the users. I'd also rather follow a tutorial that has a single repo/dir for related artifacts.

My guess is we'll need both 😅 
1. We need to provide a transition from the BYFN example since it's the defacto Fabric example nowadays. 
2. We need a clean, from-scratch example completely maintained by us. (My initial thought is that the following example design could provide a solid basis for a Caliper benchmark implementation: https://hyperledger-fabric.readthedocs.io/en/latest/developapps/scenario.html)

ericszw:
Agree.
[...]

PR #946, opened 2020-08-04
0.3.2 tutorial for Fabric Network

RosieMurphy0:
Contributes to #928 

Added a tutorial that walked through using Caliper on an existing fabric Network

Tutorial can be followed from here: https://rosiemurphy0.github.io/caliper/v0.3.2/benchmark-fabric-tutorial/


nklincoln:
replaced by #957

Issue #637, opened 2019-11-05
Provide simpler Prometheus metrics in worker clients


aklenik:
## Context
The worker clients provide some aggregated metrics related to the TXs since the last push/update. This is not the Prometheus way, as the Prometheus queries usually specify their own aggregation window.

## Expected Behavior
Specify only the following metrics:
* _caliper_tx_submitted_: counter, incremented when `submitCallback` is called, with the client id as a label.
* _caliper_tx_finished_: counter, incremented when the callback returns the result array, with the client ID and final status (success/failed) as labels
* _caliper_tx_e2e_latency_: histogram, updated when _caliper_tx_finished_ is updated (histograms are cheap on the client-side, unlike summaries)

A number of metrics may be derived from the above using Prometheus queries, like goodput, failure rate, total TPS (successful and failed), sending TPS and pending transactions. With little to no performance cost for the worker clients.

## Actual Behavior
Metrics are locally aggregated, then pushed. 

## Your Environment
* Version used: 0.2.0

PR #960, opened 2020-08-26
add prometheus scrape target tx observer

nklincoln:
Contributes to #637

Provides a prometheus-tx-observer that enables each worker to act as a scrape target with the following metrics exposed:
- caliper_tx_submitted (counter)
- caliper_tx_finished (counter)
- caliper_tx_e2e_latency (histogram)

User may provide options for:
- metrics path override
- exposed port
- processMetricCollectInterval flag
- additional labels (applied to all metrics)
- histogramBuckets override (explicit/linear/exponential)

Follow up PR should modify they PushGateway to use the same metrics


packages/caliper-core/lib/worker/tx-observers/prometheus-tx-observer.js
```diff
@@ -0,0 +1,185 @@
...
+        this.counterTxSubmitted.inc(count);
+    }
+
+    /**
+     * Called when TXs are finished.
+     * @param {TxStatus | TxStatus[]} results The result information of the finished TXs. Can be a collection of results for a batch of TXs.
+     */
+    txFinished(results) {
+        if (Array.isArray(results)) {
+            for (let result of results) {
```
davidkel, line 161: let should be const

davidkel (approved):

PR #967, opened 2020-08-27
Add support to retrieve channel creation information

davidkel:

### DIFF
--- a/packages/caliper-fabric/test/utils/GenerateConfiguration.js
+++ b/packages/caliper-fabric/test/utils/GenerateConfiguration.js
@@ -26,7 +26,6 @@ const path = require('path');
 class GenerateTestConfiguration {
 
     /**
-     *
      * @param {string} [baseConfigurationPath] a path to an base configuration to be modified
      */
     constructor(baseConfigurationPath) {
@@ -42,7 +41,7 @@ class GenerateTestConfiguration {
     }
 
     /**
-     * override the base configuration, currently only allows overriding at the root of the object
+     * Generate a configuration overriding (replacing or adding) part of the root configuration
      *
      * @param {*} overrideBaseConfiguration object which is applied to the base configuration to override
      * @returns {string} A Path to the new configuration file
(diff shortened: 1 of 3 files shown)


packages/caliper-fabric/test/utils/GenerateConfiguration.js
```diff
@@ -64,5 +79,23 @@ class GenerateTestConfiguration {
     _deepCloneObject(object) {
```
nklincoln, line 79: I think this could go into "CaliperUtils" now .. there's quite a few instances of the deep clone pattern in the code base

packages/caliper-fabric/test/utils/GenerateConfiguration.js
```diff
@@ -64,5 +79,23 @@ class GenerateTestConfiguration {
...
         return JSON.parse(JSON.stringify(object));
     }
+
+    /**
+     *
+     * @param {*} object b
+     * @param {*} propertyName b
```
nklincoln, line 86: currently the caliper code base has full descriptions in the JSDoc, would be nice to be consistent here even though the JSDoc is not published

packages/caliper-fabric/test/utils/GenerateConfiguration.js
```diff
@@ -56,6 +56,21 @@ class GenerateTestConfiguration {
         return newConfigurationFilePath;
     }
 
+    /**
+     *
+     * @param {*} propertyName b
```
nklincoln, line 61: same comment as lower down

nklincoln (changes requested):
very minor comments, but looking 🥇

packages/caliper-fabric/test/utils/GenerateConfiguration.js
davidkel: Would it make sense to raise an issue to move all of these types of calls into a single point or would you prefer to create in this PR and have the others fixed elsewhere ?

davidkel: actually just found a couple of other good 1 liners
```
let cloned = Object.assign({}, source);
```
or
```
let cloned = { ... source };
```
Think I will get rid of the method and use these instead

davidkel:
Dang, couldn't use the alternatives for cloning. JSON.parse(JSON.stringify(... is not a good generic object cloning technique, but is good enough here, so don't want to make it a general util.

nklincoln (approved):

PR #974, opened 2020-09-01
Modify prometheus monitor

nklincoln:
Contributes to #964 
- Adds basic auth
- changes to "include" option, that is regex based
- changes to array of query objects to run

Requires companion docs for breaking change

Signed-off-by: nkl199@yahoo.co.uk <nkl199@yahoo.co.uk>

Issue #964, opened 2020-08-26
Prometheus resource monitor layout is not intuitive


nklincoln:
Current configuration layout for the Prometheus resource monitor is not intuitive, it would be better if:
- [ ] we explicitly name items to capture resource stats for (regex based), under an include list
- [ ] we explicitly call out the queries to be made for the identified resource in the queries list
- [ ] we should facilitate the use of basic auth when connecting to prometheus (runtime parameters only)

The resulting monitor config file would then be:
```
monitors:
    resource:
    - module: prometheus
      options:
        interval: 5
        url: "http://localhost:9090"
        metrics:
            include: [dev-*, peer0.org2.example.com]
            queries:
                - name: Endorse Time (s)
                   query: <prom query>
                   label: instance name
                   step: 1
                   statistic: <avg | min | max>
                   multiplier: <value>

```

nklincoln:
closing due to above PRs

PR #975, opened 2020-09-01
Docs: Update resource monitor for prometheus

nklincoln:
Contributes to #964
- companion docs for #974

### DIFF
--- a/docs/vNext/MonitorsAndObservers.md
+++ b/docs/vNext/MonitorsAndObservers.md
@@ -94,14 +94,15 @@ All data stored on Prometheus may be queried by Caliper using the Prometheus que
 The prometheus monitoring module options comprise:
 - interval: monitor update interval
 - url: The Prometheus URL, used for direct queries
-- metrics: The queries to be run for inclusion within the Caliper report, comprised of to keys: `ignore` and `include`. 
-  - `ignore` a string array that is used as a blacklist for report results. Any results where the component label matches an item in the list, will *not* be included in a generated report.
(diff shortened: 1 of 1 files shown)


docs/vNext/MonitorsAndObservers.md
```diff
@@ -94,11 +94,12 @@ All data stored on Prometheus may be queried by Caliper using the Prometheus que
 The prometheus monitoring module options comprise:
 - interval: monitor update interval
 - url: The Prometheus URL, used for direct queries
-- metrics: The queries to be run for inclusion within the Caliper report, comprised of to keys: `ignore` and `include`. 
-  - `ignore` a string array that is used as a blacklist for report results. Any results where the component label matches an item in the list, will *not* be included in a generated report.
-  - `include` a series of blocks that describe the queries that are to be run at the end of each Caliper test.
+- metrics: The queries to be run for inclusion within the Caliper report, comprised of to keys: `include` and `queries`. 
+  - `include` a string array that is used to determine metric inclusion through javascript regex. Any results where the component label matches an item in the list via regex, will be included in a generated report.
```
davidkel, line 98: Sorry dumb question here, but what is a component label ?

nklincoln: no such thing as a bad question! So a "component label" is basically anything "named" that prometheus knows about, such as a docker container name. The context jumps around a little depending on the specific queries you are running and the scrape targets available but in the general case a prometheus query will come back with a set of ({componentName: value}, {componentName: value}, ...) and we can then filter on the things we are actually interested in

davidkel (approved):

PR #985, opened 2020-09-03
Update Fabric tutorial for 0.4.0

### LINKED ISSUE #962: Update Fabric tutorial for 0.4.0 (unstable)
The existing 0.3.2 tutorial requires updating for the breaking changes within 0.4.0

nklincoln:
closes #962 
- updates vNext tutorial for v0.4-unstable 

Available for view on https://nklincoln.github.io/caliper/vNext/fabric-tutorial/

### DIFF
--- a/docs/vNext/Fabric_Tut.html
+++ b/docs/vNext/Fabric_Tut.html
@@ -1,8 +1,8 @@
 ---
-layout: v0.3.2
+layout: vNext
 title:  "Fabric Tutorials"
 categories: tutorial
-permalink: /v0.3.2/fabric-tutorial/
+permalink: /vNext/fabric-tutorial/
 order: 1
 ---
 
--- a/docs/vNext/tutorials_fabric_existing.md
+++ b/docs/vNext/tutorials_fabric_existing.md
@@ -1,0 +1,631 @@
+---
+layout: vNext
+title:  "Setting up and Running a Performance Benchmark on an existing network"
+categories: 
+permalink: /vNext/fabric-tutorial/tutorials-fabric-existing/
+order: 
+---
+## Table of Contents
+{:.no_toc}
+
+- TOC
+{:toc}
+
+## Overview
+This tutorial takes you through performance testing a smart contract on a pre-existing Fabric network using Caliper. 
+
+To complete this tutorial you will need to have installed NodeJS. To do this, we recommend using [nvm](https://github.com/nvm-sh/nvm).
+
+This tutorial is based on resources available from the official [Hyperledger Fabric documentation](https://hyperledger-fabric.readthedocs.io/en/release-2.2/tutorials.html). A network comprised of two organizations and a solo orderer, with the javascript `asset-transfer-basic` smart contract, is assumed to be built and ready to performance test. 
+
+> The following command list is a minimalist quick step guide to get the required Fabric network up and running. __We use available Hyperledger Fabric resources at explicit levels. To understand and troubleshoot what occurs during the creation of the test network, please refer to the Fabric documentation linked above!__
+
+```bash
+# Clone a fixed version of the Hyperledger Fabric samples repo
+git clone https://github.com/hyperledger/fabric-samples.git
+cd fabric-samples
+git checkout 22393b629bcac7f7807cc6998aa44e06ecc77426
+# Install the Fabric tools and add them to PATH
+curl -sSL https://bit.ly/2ysbOFE | bash -s -- 2.2.0 1.4.8 -s
+export PATH=$PATH:$(pwd)/bin
+# Create and initialize the network
+cd test-network
+./network.sh up createChannel
+./network.sh deployCC -ccn basic -ccl javascript
+```
+
+## Step 1 - Create a Caliper Workspace
+Create a folder named **caliper-workspace** at the same level as the **fabric-samples** directory, and then within the **caliper-workspace** folder, create three folders named **networks**, **benchmarks**, and **workload** respectively
+
+Caliper installation and use will be based on a local npm installation. Within the **caliper-workspace** directory, install caliper CLI using the following terminal command:
+
+`npm install --only=prod @hyperledger/caliper-cli@0.4.0`
+
+Bind the SDK using the following terminal command:
+
+`npx caliper bind --caliper-bind-sut fabric:2.1`
+
+Further information relating to the installation and binding of Caliper may be found within the relevant [documentation pages](./Installing_Caliper.md).
+
+Caliper requires two configuration files:
+- The network configuration file, which describes the network under test and provides the test identities to use. 
+- The benchmark file, which defines the performance test to be completed via a sequenced set of test rounds, each specifying a workload module and a series of options to drive the workload over a time interval. 
+
+We will now populate these folders with the assets required by Caliper. 
+
+## Step 2 - Build a Network Configuration File
+The network configuration file is the file required by Caliper workers to create a connection to the existing Fabric network so that they may submit transactions. It resembles the Fabric common connection profile, augmented with additional required fields. The file can be in YAML or JSON format, this tutorial shows the JSON format. 
+
+### Create a Template Network Configuration File
+Under the **networks** folder create a template file called **networkConfig.json** with the following content:
+``` json
+{
+    "version" : "1.0",
+    "name": "Caliper test",
+    "caliper" : {
+        "blockchain": "fabric"
+    },
+    "clients" : {
+    },
+    "channels" : {
+    },
+    "organizations" : {
+    },
+    "peers" : {
+    }
+}
+```
+__version__: The version of the configuration file being used. Only "1.0" is accepted.
+
+__name__: The name for the configuration, in this instance "Caliper test".
+
+__caliper__: Indicates to Caliper the SUT that is being targeted, and may contain additional start/end commands that are not required within this tutorial. For the purposes of this tutorial, we are targeting a `fabric` network.
+
+__clients__: Lists the identities to be used within the performance test
+
+__channels__: Describes the Hyperledger Fabric channels available, their state, and the smart contracts deployed on these channels
+
+__organizations__: The Hyperledger Fabric organizations that are to be used within the benchmarking
+
+__peers__: The Hyperledger Fabric peers that are to be used within the benchmarking
+
+### Populating The Template File
+Following the test-network tutorial, a Common Connection Profile (CCP) is generated; we will use this file to help populate the Hyperledger Fabric specific elements of the Caliper network configuration file. The template will have some unique certificates for the Hyperledger Fabric network that was created, therefore you will notice differences to contents shown in subsequent stages; it is essential that you use the certificates created for interaction with the test-network you wish to target.
+
+We will be using Org1 to connect in this example. To find the generated JSON or YAML CCP files, look in **fabric-samples** -> **test-network** -> **organizations** -> **peerOrganizations** -> **org1.example.com**. We assume use of the json CCP **connection-org1.json**. 
+
+#### Organizations and Peers
+These are populated using the contents of the CCP:
+1. Copy the `organizations` and `peers` elements of the CCP into the Caliper network configuration file. 
+2. Within the `organizations.Org1` object, delete the `certificateAuthorities` listing. This is done for two reasons:
+    
+    i) We will not be using the certificate authority to register or enrol
+
+    ii) There has been a breaking change delivered to the CCP that is incompatible with the validator Caliper applies to the configuration files.
+
+Your Caliper network configuration file should now look similar to this:
+``` json
+{
+    "version" : "1.0",
+    "name": "Caliper test",
+    "caliper" : {
+        "blockchain": "fabric"
+    },
+    "clients" : {
+    },
+    "channels" : {
+    },
+    "organizations": {
+        "Org1": {
+            "mspid": "Org1MSP",
+            "peers": [
+                "peer0.org1.example.com"
+            ]
+        }
+    },
+    "peers": {
+        "peer0.org1.example.com": {
+            "url": "grpcs://localhost:7051",
+            "tlsCACerts": {
+                "pem": "-----BEGIN CERTIFICATE-----<UNIQUE CONTENT>-----END CERTIFICATE-----\n"
+            },
+            "grpcOptions": {
+                "ssl-target-name-override": "peer0.org1.example.com",
+                "hostnameOverride": "peer0.org1.example.com"
+            }
+        }
+    }
+}
+```
+
+#### Clients
+Specify the identities to be used by Caliper when performing the benchmark. Identities *must* be valid, meaning they must be known to the Fabric network and have the corresponding crypto material for use. Identities are listed within a `clients` stanza. Here we use the single identity `Admin@org1.example.com`, within which we nest the `client` object from the CCP to indicate the organization to which the identity belongs and provide basic connection timeout information. 
+
+```json
+"clients": {
+    "Admin@org1.example.com": {
+        "client": {
+            "organization": "Org1",
+            "connection": {
+                "timeout": {
+                    "peer": {
+                        "endorser": "300"
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+Under the `client` object add a property called `credentialStore` under this property add a property called `path` that has a string variable leading to a temp file in the workspace with the name `/tmp/org1`. Also under the `credentialStore` property add a property called `cryptoStore` and under this add another `path` property that points to the same temp file above, `/tmp/org1`. 
+
+This is what should be added to the `client` : 
+
+```json
+"credentialStore": {
+    "path": "/tmp/org1",
+    "cryptoStore": {
+        "path": "/tmp/org1"
+    }
+}
+```
+
+Under the `client` object add a property called `clientPrivateKey` under this property add a property called `path` that has a string variable leading to the private key of the identity. Note that the provided path is workspace relative. In this example the private key is located at **fabric-samples** -> **test-network** -> **organizations** -> **peerOrganizations** -> **org1.example.com** -> **users** -> **Admin@org1.example.com** -> **msp** -> **keystore** -> **priv_sk** .
+
+This is what should be added to the `client` object:
+
+```json
+"clientPrivateKey": {
+    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore/priv_sk"
+}
+```
+
+Also under the `client` object add another property called `clientSignedCert` and under this add a property called `path` that has a string variable leading to the signed certificate of the identity. Again note that the provided path is workspace relative. In this example it is located at **fabric-samples** -> **test-network** -> **organizations** -> **peerOrganizations** -> **org1.example.com** -> **users** -> **Admin@org1.example.com** -> **msp** -> **signedcerts** -> **admin@org1.example.com-cert.pem**
+
+This is what should be added to the `client` object:
+
+```json
+"clientSignedCert": {
+    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/signcerts/Admin@org1.example.com-cert.pem"
+}
+```
+
+The `clients` object should now look like this: 
+
+```json
+    "clients": {
+        "Admin@org1.example.com": {
+            "client": {
+                "organization": "Org1",
+                "connection": {
+                    "timeout": {
+                        "peer": {
+                            "endorser": "300"
+                        }
+                    }
+                },
+                "credentialStore": {
+                    "path": "tmp/hfc-kvs/org1",
+                    "cryptoStore": {
+                        "path": "tmp/hfc-kvs/org1"
+                    }
+                },
+                "clientPrivateKey": {
+                    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore/priv_sk"
+                },
+                "clientSignedCert": {
+                    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/signcerts/Admin@org1.example.com-cert.pem"
+                }
+            }
+        }
+    }
+```
+
+#### Channels Object
+The Fabric connector for Caliper requires assistance when creating connections to a Fabric network. A `channels` object must be provided that lists the smart contracts that may be interacted with.
+
+Add the known channel, `mychannel`, that was created by the `test-network` into the `channels` section of the Caliper network configuration file and give it the property `created` with a boolean value of true. We must list the smart contracts available on this channel as objects within an array named `contracts`. Each smart contract object has two properties, `id` and `version`. `id` specifies contract ID; in this case it is `basic`. The version is the specific contract version; in this case it is `1.0.0`. Add this as an object within the array such that the resulting `channels` object within the Caliper network configuration file becomes:
+
+```json
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"basic",
+                    "version":"1.0.0"
+                }
+            ]
+        }
+    }
+```
+
+### The Complete Network Configuration File
+The Caliper network configuration file should now be fully populated. It can be useful to take time to look over and ensure that the paths to the certificates and keys are correct. 
+
+```json
+{
+    "version" : "1.0",
+    "name": "Caliper test",
+    "caliper" : {
+        "blockchain": "fabric"
+    },
+    "clients": {
+        "Admin@org1.example.com": {
+            "client": {
+                "credentialStore": {
+                    "path": "/tmp/org1",
+                    "cryptoStore": {
+                        "path": "/tmp/org1"
+                    }
+                },
+                "organization": "Org1",
+                "clientPrivateKey": {
+                    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore/priv_sk"
+                },
+                "clientSignedCert": {
+                    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/signcerts/Admin@org1.example.com-cert.pem"
+                },
+                "connection": {
+                    "timeout": {
+                        "peer": {
+                            "endorser": "300"
+                        }
+                    }
+                }
+
+            }
+        }
+    },
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"basic",
+                    "version":"1.0.0"
+                }
+            ]
+        }
+    },
+    "organizations":{
+        "Org1": {
+            "mspid": "Org1MSP",
+            "peers": [
+                "peer0.org1.example.com"
+            ]
+        }
+    },
+    "peers": {
+        "peer0.org1.example.com": {
+            "url": "grpcs://localhost:7051",
+            "tlsCACerts": {
+                "pem": "-----BEGIN CERTIFICATE-----\n<UNIQUE CONTENT>\n-----END CERTIFICATE-----\n"
+            },
+            "grpcOptions": {
+                "ssl-target-name-override": "peer0.org1.example.com",
+                "hostnameOverride": "peer0.org1.example.com"
+            }
+        }
+    }
+}
+```
+
+## Step 3 - Build a Test Workload Module
+The workload module interacts with the deployed smart contract during the benchmark round. The workload module extends the Caliper class `WorkloadModuleBase` from `caliper-core`. The workload module provides three overrides: 
+- `initializeWorkloadModule` - used to initialize any required items for the benchmark
+- `submitTransaction` - used to interact with the smart contract method during the monitored phase of the benchmark
+- `cleanupWorkloadModule` - used to clean up after the completion of the benchmark
+
+For more information, please see the specific documentation on `Workload Configuration` accessible on the left hand menu.
+
+The workload we will be driving aims to benchmark the querying of existing assets within the world state database. Consequently we will use all three phases available in the workload module:
+- `initializeWorkloadModule` - to create assets that may be queried in the `submitTransaction` phase
+- `submitTransaction` - to query assets created in the `initializeWorkloadModule` phase
+- `cleanupWorkloadModule` - used to remove assets created in the `initializeWorkloadModule` phase so that the benchmark may be repeated
+
+### Create A Template Workload Module
+Within the **workload** folder create a file called **readAsset.js** with the following content: 
+
+``` js
+'use strict';
+
+const { WorkloadModuleBase } = require('@hyperledger/caliper-core');
+
+class MyWorkload extends WorkloadModuleBase {
+    constructor() {
+        super();
+    }
+    
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+    }
+    
+    async submitTransaction() {
+        // NOOP
+    }
+    
+    async cleanupWorkloadModule() {
+        // NOOP
+    }
+}
+
+function createWorkloadModule() {
+    return new MyWorkload();
+}
+
+module.exports.createWorkloadModule = createWorkloadModule;
+```
+
+### Populating the Template File
+When we populate this file we make reference to the available methods within the deployed smart contract **assetTransfer.js** file that can be found in: **fabric-samples** -> **asset-transfer-basic** -> **chaincode-javascript** -> **lib** -> **assetTransfer.js**
+
+#### Populate initializeWorkloadModule
+This method is used to prepare any items required by the primary `submitTransaction` method as the benchmark completes. 
+
+The number of assets to be created will be given as `roundArguments.assets`. We create assets using the smart contract by populating an arguments object, which defines the transaction body, and using the Caliper API `sendRequests`, which requires knowledge of:
+- contractId, the name of smart contract that is to be used and is present within the Caliper network configuration file
+- contractFunction, the specific function within the smart contract to invoke
+- contractArguments, the arguments to pass to the smart contract function
+- invokerIdentity, the identity to use that is present within the Caliper network configuration file
+- readOnly, if performing a query operation or not
+
+The method should look like this:
+```js
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+
+        for (let i=0; i<this.roundArguments.assets; i++) {
+            const assetID = `${this.workerIndex}_${i}`;
+            console.log(`Worker ${this.workerIndex}: Creating asset ${assetID}`);
+            const request = {
+                contractId: this.roundArguments.contractId,
+                contractFunction: 'CreateAsset',
+                invokerIdentity: 'Admin@org1.example.com',
+                contractArguments: [assetID,'blue','20','penguin','500'],
+                readOnly: false
+            };
+
+            await this.sutAdapter.sendRequests(request);
+        }
+    }
+```
+
+In the above example, different assets will be created that have the same parameters (blue, 20, penguin, 500). Comparing the above to the smart contract method itself, it should be evident that there is a 1:1 mapping of contract arguments to the method parameters.
+
+#### Populate submitTransaction
+This method runs repeatedly in the benchmark test phase. We will be evaluating the `ReadAsset` smart contract method by querying the assets we created in the `initializeWorkloadModule` method. 
+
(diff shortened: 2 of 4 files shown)


docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+## Step 1 - Install and Bind Caliper
+
+Globally install caliper CLI using the following terminal command:
+
+`npm install -g --only=prod @hyperledger/caliper-cli@0.4.0`
+
+Bind the SDK using the following terminal command:
+
+`caliper bind --caliper-bind-sut fabric:2.2.0 --caliper-bind-args=-g`
```
davidkel, line 29: should we use fabric:2.2 rather than 2.2.0 ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+The network configuration file is the file required by Caliper workers to create a connection to the existing Fabric network so that they may submit transactions. It resembles the Fabric common connection profile, augmented with additional required fields. The file can be in YAML or JSON format, this tutorial shows the JSON format. 
+
+### Create The File
+
+Under the **networks** folder create a file called **network_config.json**.
+
+The file is made up of eight objects: version, caliper, clients, channels, name, organization, peers, and certificateAuthorities. Each of these we will be populated in subsequent steps through modification of a template file output by the Hyperledger Fabric test-network tutorial. 
+
+``` bash
```
davidkel, line 51: Might be useful to explain what this example represents, because it's not actually used.

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+We will now populate these folders with the assets required by Caliper. 
+
+## Step 3 - Network Configuration File
+
+The network configuration file is the file required by Caliper workers to create a connection to the existing Fabric network so that they may submit transactions. It resembles the Fabric common connection profile, augmented with additional required fields. The file can be in YAML or JSON format, this tutorial shows the JSON format. 
+
+### Create The File
+
+Under the **networks** folder create a file called **network_config.json**.
```
davidkel, line 47: Why do we need to create this file now ? Further down the tutorial we could just copy the connection profile rather than cut and paste to overwrite any contents in hnetwork_config.json

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+                        "endorser": "300"
+                    }
+                }
+            }
+        }
+    }
+},
+```
+
+Under the `client` object add a property called `credentialStore` under this property add a property called `path` that has a string variable leading to a temp file in the workspace with the name `tmp/hfc-kvs/org1`. Also under the `credentialStore` property add a property called `cryptoStore` and under this add another `path` property that points to the same temp file above, `tmp/hfc-kvs/org1`.
```
davidkel, line 170: Need a comma or fullstop before the 2nd `under` in this statement
The tmp/hfc-kvs/org1, I think are directories rather than temp files. Also should we specify /tmp rather than tmp ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+`caliper bind --caliper-bind-sut fabric:2.2.0 --caliper-bind-args=-g`
+
+## Step 2 - Setting Up Folders
+
+Caliper needs two configuration files:
+- The network configuration file that describes the network and provides the connection requirements. 
+- The benchmark file that defines the benchmark test rounds, through specification of a workload module and a series of options to drive the workload over a time interval. 
+
+Create a parent folder named **caliper-workspace** and then within this folder create two subfolders named **networks** and **benchmarks** respectively
```
davidkel, line 37: the tutorial assumes the caliper-workspace and fabric-samples are at the same dir level, such that `../fabric-samples` would be valid. Might be worth noting

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+The workload we will be driving aims to benchmark the querying of existing cars within the world state database. Consequently we will use all three phases available in the workload module:
+- `initializeWorkloadModule` - to create cars that may be queried in the `submitTransaction` phase
+- `submitTransaction` - to query cars created in the `initializeWorkloadModule` phase
+- `cleanupWorkloadModule` - used to remove cars created in the `initializeWorkloadModule` phase so that the benchmark may be repeated
+
+### Create The Workload Module
+
+Under the **benchmarks** folder create a subfolder called **workload**, and navigate to this folder.
+1. Run `npm init` to create a `package.json` file. Open the file and add `"@hyperledger/caliper-core": "0.4.0"` as a dependency.
```
davidkel, line 427: maybe give an example of a package.json as it will be small

davidkel, line 427: maybe use npm instead and just do
```
npm install @hyperledger/caliper-core@0.4.0 --save
``` 
?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+const { WorkloadModuleInterface } = require('@hyperledger/caliper-core');
+
+class MyWorkload extends WorkloadModuleInterface {
+    constructor() {
+        super();
+    }
+    
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+        this.contractID = 'fabcar';
```
davidkel, line 552: this.contractID and this.version are not required

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+The whole file should look like this:
+
+```bash
+'use strict';
+
+const { WorkloadModuleInterface } = require('@hyperledger/caliper-core');
+
+class MyWorkload extends WorkloadModuleInterface {
+    constructor() {
```
davidkel, line 546: guess we could drop the constructor definition

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+function createWorkloadModule() {
+    return new MyWorkload();
+}
+
+module.exports.createWorkloadModule = createWorkloadModule;
+```
+
+## Step 5 - Benchmark Configuration File
+
+The benchmark configuration file defines the benchmark rounds and references the defined callbacks. It will specify the number of test workers to use when generating the load, the number of test rounds, the duration of each round, the rate control applied to the transaction load during each round through execution of a workload module, and options relating to monitors. This particular tutorial will not make use of any of the available monitors or transaction observers; for these details please refer to the documentation.
```
davidkel, line 604: Should `callbacks` now be `workloads` ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+First add a root level block named `test`. In this block add:
+
+- a `name` key with the value `my-asset-benchmark`
+- a `description` key with a short description as the value, in this example that is `test benchmark`
+- a literal block named `workers` that has the following keys and values:
+    - `type: local`
+    - `number: 2`
+
+It should look like this so far:
```
davidkel, line 627: But nothing follows this statement of what it should look like

docs/vNext/tutorials_fabric_existing.md
davidkel: Also the node-sdk has a slight problem which means we can't use any of the 2.2 sdks at the moment because fabric-common forgets to declare it's dependency on long so loading fabric-network fails because long is missing. This I think only happens because the node-sdk modules are loaded into global and node.js require doesn't really do loading modules from global with explicitly putting it on the path. I am assuming it is put into global so caliper-cli will pick it up as it's installed globally as well ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+- `caliper-fabric-gateway-enabled`
+- `caliper-fabric-gateway-discovery`
+
+### Run the command
+
+Ensure that you are in the caliper-workspace directory. 
+
+In the terminal run the following Caliper CLI command:
+
+`caliper launch manager --caliper-benchconfig benchmarks/myAssetBenchmark.yaml --caliper-networkconfig networks/network_config.json --caliper-workspace ./ --caliper-flow-only-test --caliper-fabric-gateway-enabled --caliper-fabric-gateway-discovery`
```
davidkel, line 709: This command failed for 2 reasons
1. long couldn't be found
2. The connection profile from fabric now confirms to a new standard and the configValidator fails. Long story but now tlsCACerts supports arrays as well as strings and the CA definition from test-network uses that format

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+    rounds:
+      - label: queryAsset
+        description: Query asset benchmark
+        txDuration: 30
+        rateControl: 
+          type: fixed-load
+          opts:
+            transactionLoad: 2
+        workload:
+          module: benchmarks/workload/queryAssetBenchmarks.js
```
davidkel, line 679: this should be queryAssetBenchmark.js

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+const { WorkloadModuleInterface } = require('@hyperledger/caliper-core');
+
+class MyWorkload extends WorkloadModuleInterface {
+    constructor() {
+        super();
+    }
+    
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
```
davidkel, line 446: need to remove this line otherwise you get an error saying you need to implement `initializeWorkloadModule`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+### initializeWorkloadModule
+
+This method is used to prepare any items required by the `submitTransaction` method as the benchmark completes. 
+
+The number of assets to be created will be given as `roundArguments.assets`. We create assets using the smart contract by populating an arguments object, which defines the transaction body, and using the Caliper API `sendRequests`. Note the `readOnly` value passed within the arguments.
+
+The method should look like this:
+```bash
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
```
davidkel, line 474: need to remove this line and also store the relevant parameters as properties of the workload object, eg
```
this.roundArguments = roundArguments
```

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+'use strict';
+
+const { WorkloadModuleInterface } = require('@hyperledger/caliper-core');
+
+class MyWorkload extends WorkloadModuleInterface {
+    constructor() {
+        super();
+    }
+    
+    async initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext) {
```
davidkel, line 550: see above comments around this method

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+        }
+    }
+    
+    async submitTransaction() {
+        const randomId = Math.floor(Math.random()*this.roundArguments.assets);
+        const myArgs = {
+            contractID: this.roundArguments.contractID,
+            contractFunction: 'queryCar',
+            invokerIdentity: 'Admin@org1.example.com',
+            contractArguments: [`${clientIdx}_${randomId}`],
```
davidkel, line 583: clientIdx isn't defined, maybe it should be this.workerIndex ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+The function should look like this:
+```bash
+    async submitTransaction() {
+        const randomId = Math.floor(Math.random()*this.roundArguments.assets);
+        const myArgs = {
+            contractID: this.roundArguments.contractID,
+            contractFunction: 'queryCar',
+            invokerIdentity: 'Admin@org1.example.com',
+            contractArguments: [`${clientIdx}_${randomId}`],
```
davidkel, line 517: clientIdx isn't defined, maybe it should be this.workerIndex ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+
+            }
+        }
+    },
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"fabcar",
```
davidkel, line 359: test-network might not be deploying fabcar as fabcar anymore, I just tried it and it was called `basic`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+``` bash
+"created" : true,
+```
+
+`contracts` holds an array with two properties, `id` and `version`. `id` specifies contract ID; in this case it is `fabcar`. The version is that specific contract version; in this case it is `0.0.1`. It should look like this: 
+
+``` bash
+"contracts": [
+    {
+        "id":"fabcar",
```
davidkel, line 252: Don't think fabcar exists anymore in the latest test-network

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+        "version":"0.0.1"
```
davidkel, line 253: Is version required if we aren't creating ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+        this.contractID = 'fabcar';
+        this.version = '0.0.1';
+
+        const requests = [];
+        for (let i=0; i<this.roundArguments.assets; i++) {
+            const assetID = `${this.workerIndex}_${i}`;
+            console.log(`Client ${this.workerIndex}: Creating asset ${assetID}`);
+            const args = {
+                contractID: this.roundArguments.contractID,
```
davidkel, line 560: property name expected by the connector is `contractId`. May we should also change the roundArguments version as well ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,724 @@
...
+        txDuration: 30
+        rateControl: 
+          type: fixed-load
+          opts:
+            transactionLoad: 2
+        workload:
+          module: benchmarks/workload/queryAssetBenchmarks.js
+          arguments:
+            assets: 10
+            contractID: fabcar
```
davidkel, line 682: contractId maybe to match the parameter expected by the connector ? Also no fabcar anymore

davidkel (changes requested):
Sorry, quite a few things to change here

docs/vNext/tutorials_fabric_existing.md
aklenik: Requiring global packages (e.g., caliper-core from the workload module) can be enabled with the following env var: https://github.com/hyperledger/caliper/blob/master/packages/caliper-publish/caliper.Dockerfile#L35-L37

docs/vNext/tutorials_fabric_existing.md
aklenik: So far I personally specified /tmp, and it was cleaned up with the end script. It could also point to the workplace-relative tmp dir, but it also needs to be cleaned up when the crypto artifacts change.

docs/vNext/tutorials_fabric_existing.md
aklenik: If the CLI is installed first, won't the global caliper-core package be picked up automatically (provided that NODE_PATH is set)?

docs/vNext/tutorials_fabric_existing.md
aklenik: ❗  The WorkloadModuleInterface doesn't provide any base implementation, the WorkloadModule base class should be used: https://github.com/hyperledger/caliper/blob/master/packages/caliper-core/lib/worker/workload/workloadModuleBase.js

docs/vNext/tutorials_fabric_existing.md
aklenik: I think it's part of the TX proposal, but not sure what happens if you omit it.

docs/vNext/tutorials_fabric_existing.md
davidkel: A fix I submitted has gone into the node-sdk, however I've changed the SUT bindings for 2.2 so we don't explicitly load protos and common, which means that we avoid the bug in the node-sdk

docs/vNext/tutorials_fabric_existing.md
davidkel: It will only be picked up automatically is node_path is set otherwise not. This appears to be by design by the node team (ie it doesn't such global by default) so my personal preference would be to keep things local.

nklincoln: yup, all caliper calls should be `npx caliper`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+{:.no_toc}
+
+- TOC
+{:toc}
+
+## Overview
+This tutorial takes you through performance testing a smart contract on a pre-existing Fabric network using Caliper. 
+
+This tutorial follows from the test-network [fabric tutorial](https://hlf.readthedocs.io/en/latest/test_network.html). You need to have installed and instantiated the `asset-transfer-basic` smart contract (part two of the Fabric tutorial).
+
```
davidkel, line 18: Might be worth mentioning that a version of node is required for this tutorial, plus is you don't want to install Go then you can install the node version of asset-transfer-basic using the command `./network.sh deployCC -ccn basic -ccl javascript`.

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+```json
+"credentialStore": {
+    "path": "/tmp/org1",
+    "cryptoStore": {
+        "path": "/tmp/org1"
+    }
+}
+```
+
+Under the `client` object add a property called `clientPrivateKey` under this property add a property called `path` that has a string variable leading to a the identities private key. Note that the provided path is workspace relative. In this example the private key is located at **fabric-samples** -> **test-network** -> **organizations** -> **peerOrganizations** -> **org1.example.com** -> **users** -> **Admin@org1.example.com** -> **msp** -> **keystore** -> **priv_sk** .
```
davidkel, line 154: test says `leading to a the identities` so either `a` or `the`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+This is what should be added to the `client` object:
+
+```json
+"clientPrivateKey": {
+    "path": "../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore/priv_sk"
+}
+```
+
+Also under the `client` object add another property called `clientSignedCert` and under this add a property called `path` that has a string variable leading to the identities signed certificate. Again note that the provided path is workspace relative. In this example it is located at **fabric-samples** -> **test-network** -> **organizations** -> **peerOrganizations** -> **org1.example.com** -> **users** -> **Admin@org1.example.com** -> **msp** -> **signedcerts** -> **admin@org1.example.com-cert.pem**
```
davidkel, line 164: same comment here about `leading to the identities`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+                    "id":"asset-transfer-basic",
+                    "version":"1.0.0"
+                }
+            ]
+        }
+    }
+```
+
+### The Complete Network Configuration File
+The Caliper network configuration file should now be fully populated. It can be useful to take time to look over and ensure that the paths to the certificate and keys are correct.
```
davidkel, line 225: certificate --> certificates

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+```
+
+In the above example, different assets will be created that have the same parameters (blue, 20, penguin, 500). Comparing the above to the smart contract method itself, it should be evident that there is a 1:1 mapping of contract arguments to the method parameters.
+
+#### Populate submitTransaction
+This method runs repeatedly in the benchmark test phase. We will be evaluating the `ReadAsset` smart contract method by querying the assets we created in the `initializeWorkloadModule` method. 
+
+First, create a string identity for the asset to query, formed by the concatenation of the worker index and a random integer between 0 and the number of created assets. 
+
+Then await the call on `sendRequests`, passing the contractId, contractFunction` set as `queryCar`; `invokerIdentity` set as `admin@org1.example.com`; and `chaincodeArguments` set as an array that contains the asset to query in this run.
```
davidkel, line 382: Something has gone wrong with the quoting on this line, plus `queryCar` isn't the correct contractFunction now

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+#### Populate Rounds
+Each `round` block contains the following: 
+- `label` - the unique header label to use for the round.
+- `description` - a description of the round being run.
+- `txDuration` - the specification of the test duration, in seconds
+- `rateControl` - a rate control type, with options. 
+- `workload` - the workload module to use, with arguments to pass to the module. All arguments passed are available as `roundArguments` within the workload module.
+
+We will specify a benchmark round labeled `queryAsset`, with the description `Query asset benchmark`, to run for a 30s duration, using a `fixed-load` rate controller aiming to maintain a constant transaction pressure of 2. Additionally we will be providing a workload through specification of our `queryAssetBenchmarks.js` workload file, which we will pass the arguments `{assets: 10, contractId: asset-transfer-basic}`.
```
davidkel, line 541: queryAsset and references to query should be changed to readAsset. The workload module is called `readAsset.js` not `queryAssetBenchmarks.js`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+The above is accomplished through the round specification:
+
+``` yaml
+    - label: queryAsset
```
davidkel, line 546: Suggest changing the label

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+      description: Query asset benchmark
```
davidkel, line 547: suggest changing the description

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+      txDuration: 30
+      rateControl: 
+        type: fixed-load
+        opts:
+          transactionLoad: 2
+      workload:
+        module: workload/queryAssetBenchmarks.js
```
davidkel, line 554: workload/readAsset.js

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+```yaml
+test:
+    name: asset-transfer-basic-benchmark
+    description: test benchmark
+    workers:
+      type: local
+      number: 2
+    rounds:
+      - label: queryAsset
```
davidkel, line 571: suggest change label, description

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+        description: Query asset benchmark
+        txDuration: 30
+        rateControl: 
+          type: fixed-load
+          opts:
+            transactionLoad: 2
+        workload:
+          module: workload/queryAssetBenchmarks.js
```
davidkel, line 579: module should be workload/readAsset.js

docs/vNext/tutorials_fabric_existing.md
davidkel: clientIdx still not defined :-)

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+Add the known channel, `mychannel`, that was created by the `test-network` into the `channels` section of the Caliper network configuration file and give it the property `created` with a boolean value of true. We must list the smart contracts available on this channel as objects within an array named `contracts`. Each smart contract object has two properties, `id` and `version`. `id` specifies contract ID; in this case it is `asset-transfer-basic`. The version is the specific contract version; in this case it is `1.0.0`. Add this as an object within the array such that the resulting `channels` object within the Caliper network configuration file becomes:
+
+```json
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"asset-transfer-basic",
```
davidkel, line 216: For me when using node the id was just `basic`. Would this be different for other chaincode languages like `Go` ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+
+            }
+        }
+    },
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"asset-transfer-basic",
```
davidkel, line 266: see above comment about the chaincode id being just `basic`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+      txDuration: 30
+      rateControl: 
+        type: fixed-load
+        opts:
+          transactionLoad: 2
+      workload:
+        module: workload/queryAssetBenchmarks.js
+        arguments:
+          assets: 10
+          contractId: asset-transfer-basic
```
davidkel, line 557: See comment above about the contract id, but I think this should be called `basic`

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+        txDuration: 30
+        rateControl: 
+          type: fixed-load
+          opts:
+            transactionLoad: 2
+        workload:
+          module: workload/queryAssetBenchmarks.js
+          arguments:
+            assets: 10
+            contractId: asset-transfer-basic
```
davidkel, line 582: contractId is `basic` ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+## Table of Contents
+{:.no_toc}
+
+- TOC
+{:toc}
+
+## Overview
+This tutorial takes you through performance testing a smart contract on a pre-existing Fabric network using Caliper. 
+
+This tutorial follows from the test-network [fabric tutorial](https://hlf.readthedocs.io/en/latest/test_network.html). You need to have installed and instantiated the `asset-transfer-basic` smart contract (part two of the Fabric tutorial).
```
aklenik, line 17: I'd add a quick command summary for getting the Fabric network to the required state. This way we can be sure that the network is set up correctly by the user. 
Strictly with the following disclaimer:
> The below command list is a minimalist quick step guide to get the Fabric network up and running. __To understand and troubleshoot them, refer to the Fabric documentation!__

```bash
# clone a fixed version of the samples repo
git clone https://github.com/hyperledger/fabric-samples.git
cd fabric-samples
git checkout v2.1.1
# install the Fabric tools and add them to PATH
curl -sSL https://bit.ly/2ysbOFE | bash -s -- 2.2.0 1.4.8 -s
export PATH=$PATH:$(pwd)/bin
# create and initialize the network
cd test-network
./network.sh up createChannel
./network.sh deployCC -l javascript
```

docs/vNext/tutorials_fabric_existing.md
aklenik: I'd also suggest using the javascript chaincode, since packaging it doesn't require 3rd party tools (and the ccenv image has every other dependency). While `go mod` requires a newer go version.

docs/vNext/tutorials_fabric_existing.md
aklenik: For the `v2.1.1` samples repo and the JS CC, I see the `fabcar` name in the script invocations :game_die: 
Btw, does the CC version have any significance beside CC installation in the old-style CC deploy process?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,611 @@
...
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+
+        for (let i=0; i<this.roundArguments.assets; i++) {
+            const assetID = `${this.workerIndex}_${i}`;
+            console.log(`Worker ${this.workerIndex}: Creating asset ${assetID}`);
+            const request = {
+                contractId: this.roundArguments.contractId,
+                contractFunction: 'CreateAsset',
+                invokerIdentity: 'Admin@org1.example.com',
+                contractArguments: [assetID,'blue',20,'penguin',500],
```
aklenik, line 366: All arguments must be string, IIRC. So `20` -> `"20"`, etc.

docs/vNext/tutorials_fabric_existing.md
aklenik: Also must prefix it with `./` so the module loader knows it's a local file and must be resolved.

docs/vNext/tutorials_fabric_existing.md
nklincoln: removed the ca aspect from the tutorial as the v2 connector when using a gateway wont need it

docs/vNext/tutorials_fabric_existing.md
nklincoln: yes 🤦

docs/vNext/tutorials_fabric_existing.md
nklincoln: I have a brain, I promise 🧠

docs/vNext/tutorials_fabric_existing.md
nklincoln: yeah, the symmetry there is good 👍

docs/vNext/tutorials_fabric_existing.md
nklincoln: added, along with a code block fast start snippet

docs/vNext/tutorials_fabric_existing.md
nklincoln: yeah, looks like i missed a backtick - modified the paragraph to try and make it more clear

docs/vNext/tutorials_fabric_existing.md
nklincoln: added the code block 🦖

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,629 @@
...
+
+To complete this tutorial you will need to have installed NodeJS. To do this, we recommend using [nvm](https://github.com/nvm-sh/nvm).
+
+This tutorial follows from the test-network [fabric tutorial](https://hyperledger-fabric.readthedocs.io/en/release-2.2/tutorials.html). You need to have installed and instantiated the javascript `asset-transfer-basic` smart contract using the Fabric tutorial command `./network.sh deployCC -ccn basic -ccl javascript` (part two of the Fabric tutorial).
+
+> The following command list is a minimalist quick step guide to get the Fabric network up and running. __To understand and troubleshoot them, refer to the Fabric documentation linked above!__
+
+```bash
+# Clone a fixed version of the Hyperledger Fabric samples repo
+git clone https://github.com/hyperledger/fabric-samples.git
```
davidkel, line 25: Unfortunately these instructions aren't correct. I think they should be removed and we should stick to what is described in the fabric tutorial itself with the option to explain how to install the node version of the basic chaincode rather than the Go version as a suggestion before pointing to the tutorial

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,629 @@
...
+#### Organizations and Peers
+These are populated using the contents of the CCP:
+1. Copy the `organizations` and `peers` elements of the CCP into the Caliper network configuration file. 
+2. Within the `organizations.Org1` object, delete the `certificateAuthorities` listing. This is done for two reasons:
+    
+    i) We will not be using the certificate authority to register or enrol
+
+    ii) There has been a breaking change delivered to the CCP that is incompatible with the validator Caliper applies to the configuration files.
+
+Your Caliper network configuration file should now look like this:
```
davidkel, line 104: should noew look `similar` to this as the <UNIQUE CONTENT> stanza may not be obvious to some

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,629 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+Under the `client` object add a property called `credentialStore` under this property add a property called `path` that has a string variable leading to a temp file in the workspace with the name `/tmp/org1`. Also under the `credentialStore` property add a property called `cryptoStore` and under this add another `path` property that points to the same temp file above, `/tmp/org1`. 
+
+This is what should be add to the `client` :
```
davidkel, line 161: add --> added

docs/vNext/tutorials_fabric_existing.md
aklenik: 1. True, the checkout needs a newer tag/commit, but we should choose a commit, so the tutorial won't break after some changes in the samples (current master is `22393b629bcac7f7807cc6998aa44e06ecc77426`). 
2. The curl line could also clone the repo, but since we want a specific version, I extracted the cloning part. 
3. The last lines are from the Fabric tutorial, but since it's a bit wordy, I extracted the gist, so users can focus on the Caliper part. 
4. We should reference the `Installing and Running Caliper/Pre-requisites` doc section for Caliper deps.

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,629 @@
...
+        await super.initializeWorkloadModule(workerIndex, totalWorkers, roundIndex, roundArguments, sutAdapter, sutContext);
+
+        for (let i=0; i<this.roundArguments.assets; i++) {
+            const assetID = `${this.workerIndex}_${i}`;
+            console.log(`Worker ${this.workerIndex}: Creating asset ${assetID}`);
+            const request = {
+                contractId: this.roundArguments.contractId,
+                contractFunction: 'CreateAsset',
+                invokerIdentity: 'Admin@org1.example.com',
+                contractArguments: [assetID,'blue',20,'penguin',500],
```
aklenik, line 384: Some args are still numbers instead of strings.

docs/vNext/tutorials_fabric_existing.md
davidkel: Version isn't required for the tx proposal as you can't select different versions of an instantiated contract on a channel. It would be useful for operational activities of install and instantiate though.

aklenik (approved):
I think the content and code is correct and user friendly, so let's see what the community thinks about it 🥇

docs/vNext/Fabric_Tut.html
```diff
@@ -0,0 +1,18 @@
...
+<p>These tutorials are specifically aimed at using Caliper to performance test Hyperledger Fabric networks.</p>
+
+<div class="section-3">
+    <div class="tut-section">
+        <a class="tut-specific" href="./tutorials-fabric-existing">
+          <strong>Running Caliper on an existing Fabric network</strong>
+          <p>This tutorial takes you through performance testing a smart contract on a pre-existing Fabric network using Caliper. </p>
+        </a>
+    </div>
+</div>
```
davidkel, line 18: missing a new line

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,631 @@
...
+Add the known channel, `mychannel`, that was created by the `test-network` into the `channels` section of the Caliper network configuration file and give it the property `created` with a boolean value of true. We must list the smart contracts available on this channel as objects within an array named `contracts`. Each smart contract object has two properties, `id` and `version`. `id` specifies contract ID; in this case it is `basic`. The version is the specific contract version; in this case it is `1.0.0`. Add this as an object within the array such that the resulting `channels` object within the Caliper network configuration file becomes:
+
+```json
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"basic",
+                    "version":"1.0.0"
```
davidkel, line 237: still keeping the version ?

docs/vNext/tutorials_fabric_existing.md
```diff
@@ -0,0 +1,631 @@
...
+            }
+        }
+    },
+    "channels": {
+        "mychannel": {
+            "created" : true,
+            "contracts": [
+                {
+                    "id":"basic",
+                    "version":"1.0.0"
```
davidkel, line 287: still keeping the version ?

davidkel (approved):
LGTM

Issue #940, opened 2020-07-30
As a Caliper user, I would like Fabric network configuration files to be more understandable, so that I can create them with greater ease


nklincoln:
Caliper network configuration files can be very confusing, which is a result of how they have evolved over time. We need to consider that the file is used by Caliper, and should consequently should be shaped in a way that suits the use within Caliper rather than being beholding to the target connector.

Configuration files for the Fabric connector should:

- [x] Point to a connection profile, instead of extending one
- [x] Specify identities to use when testing, which may be one, or a combination of:
  - [x] One or more (file) wallets, from which identities are used, without having to nest a larger client description
  - [x] An identity backed with existing certificates
  - [x] An identity name to be registered/enrolled, using available registrar information (DROPPED)
- [x] Indicate if the network being targeted has discovery enabled or not
- [x] Not require unnecessary administrative certificates if not explicitly required

It would be nice if the above considerations could be introduced, such that it is more obvious how the network configuration file is constructed, and subsequently consumed by caliper.

davidkel:
![image](https://user-images.githubusercontent.com/25582377/91459722-dc6ee580-e87e-11ea-9a5c-8b70ee0d45de.png)

davidkel:
Proposed new format for fabric adapter configuration (caveat, may still have minor iterative changes as this issue is progressed)
```
---
# The name of the benchmark
name: Geoff
# The caliper semantic version
version: "1.0.0"

# Global properties that define the SUT, and options that are specific to the target SUT
caliper:
    blockchain: fabric,
    sutOptions :
       mutualTls: true

# Array of fabric channels for creation or use
channels:
    - channelName: my-channel
      # [Optional] create the channel
      create: true
      definition:
        capabilities: []
        consortium: 'SampleConsortium2'
        msps: ['Org1MSP', 'Org2MSP']
        version: 0
        orderers:
        - orderer0.example.com
        - orderer1.example.com
        peers:
            peer0.org1.example.com:
                eventSource: true
            peer0.org2.example.com:
                eventSource: true
      # Array of contracts to be installed/instantiated on the named channel and available for use by the workload module
      contracts:
      - id: marbles
        contractID: myMarbles
        version: v0
        language: golang
        path: marbles/go
        metadataPath: src/marbles/go/metadata

    - channelName: your-channel
      create: false
      contracts:
      - id: marbles
        contractID: lostMyMarbles
        version: v0

[...]

PR #991, opened 2020-09-08
added getter for contract definitions

RosieMurphy0:
Signed-off-by: RosieMurphy0 <rosie.murphy@ibm.com>

Adds a function that returns the array that holds the contracts for a specified channel from the config file. 

Contributes to #940

### DIFF
--- a/packages/caliper-fabric/lib/ConnectorConfiguration.js
+++ b/packages/caliper-fabric/lib/ConnectorConfiguration.js
@@ -86,6 +86,17 @@ class ConnectorConfiguration {
 
         return null;
     }
+    /**
+     * return array of contract definitions for specific channel
+     * @param {string} channelPassed channel wanted
+     * @returns {Array} all of contract definitions for channel
+    */
+    getContractDefinitionsForChannelName(channelPassed) {
+        const channelList = this.adapterConfiguration.channels;
+        const channelWanted = channelList.filter(channelContracts => channelContracts.channelName === channelPassed);
+        const contractDefinitions = channelWanted[0].contracts;
+        return contractDefinitions;
+    }
 }
 
 module.exports = ConnectorConfiguration;
--- a/packages/caliper-fabric/lib/ConnectorConfiguration.js
+++ b/packages/caliper-fabric/lib/ConnectorConfiguration.js
@@ -88,14 +88,19 @@ class ConnectorConfiguration {
     }
     /**
      * return array of contract definitions for specific channel
-     * @param {string} channelPassed channel wanted
+     * @param {string} channelName channel name wanted
      * @returns {Array} all of contract definitions for channel
     */
-    getContractDefinitionsForChannelName(channelPassed) {
+    getContractDefinitionsForChannelName(channelName) {
         const channelList = this.adapterConfiguration.channels;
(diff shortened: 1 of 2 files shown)


packages/caliper-fabric/lib/ConnectorConfiguration.js
```diff
@@ -86,6 +86,17 @@ class ConnectorConfiguration {
...
         return null;
     }
+    /**
+     * return array of contract definitions for specific channel
+     * @param {string} channelPassed channel wanted
+     * @returns {Array} all of contract definitions for channel
+    */
+    getContractDefinitionsForChannelName(channelPassed) {
+        const channelList = this.adapterConfiguration.channels;
+        const channelWanted = channelList.filter(channelContracts => channelContracts.channelName === channelPassed);
```
davidkel, line 96: As we don't have a validator yet and not sure when a validator would be made available for now lets add some defensive checks here to make sure that there is a channelList, so you can check the example used in getChannelNamesForCreation.

packages/caliper-fabric/lib/ConnectorConfiguration.js
```diff
@@ -86,6 +86,17 @@ class ConnectorConfiguration {
 
         return null;
     }
+    /**
+     * return array of contract definitions for specific channel
+     * @param {string} channelPassed channel wanted
+     * @returns {Array} all of contract definitions for channel
+    */
+    getContractDefinitionsForChannelName(channelPassed) {
```
davidkel, line 94: would suggest changing `channelPassed` to `channelName` for the parameter

packages/caliper-fabric/lib/ConnectorConfiguration.js
```diff
@@ -86,6 +86,17 @@ class ConnectorConfiguration {
...
+        const channelList = this.adapterConfiguration.channels;
+        const channelWanted = channelList.filter(channelContracts => channelContracts.channelName === channelPassed);
+        const contractDefinitions = channelWanted[0].contracts;
```
davidkel, line 97: For consistency, if the contracts definition is not there, then we should just return an empty array. Currently if it wasn't present we would be returning null or undefined. This will need a further test to check for.

packages/caliper-fabric/lib/ConnectorConfiguration.js
```diff
@@ -86,6 +86,22 @@ class ConnectorConfiguration {
...
+     * return array of contract definitions for specific channel
+     * @param {string} channelName channel name wanted
+     * @returns {Array} all of contract definitions for channel
+    */
+    getContractDefinitionsForChannelName(channelName) {
+        const channelList = this.adapterConfiguration.channels;
+
+        if (channelList && Array.isArray(channelList)){
+            const channelWanted = channelList.filter(channelContracts => channelContracts.channelName === channelName);
+            const contractDefinitions = channelWanted[0].contracts;
```
davidkel, line 99: Should test here what is in contracts, eg
```
if (contractDefinitions && 
    Array.isArray(contractDefinitions) {
   return contractDefinitions;
}
```
Then add a test to see that you get an empty array is the contracts property hasn't been specified

davidkel (approved):
LGTM

PR #993, opened 2020-09-09
fix SUT and error reporting

### LINKED ISSUE #988: Fabric SUT 2.2's fail to load when globally bound
There are a couple of issues here 
The binding definitions 2.2 are all like this (taken from the way 2.1 was done)
```
 packages: ['fabric-common@2.2.2', 'fabric-protos@2.2.2', 'fabric-network@2.2.2', 'fabric-ca-client@2.2.2']
```
The problem here is that due to a bug in the node-sdk 2.2 fabric-common fails to load because it can't find `long`.

So the first problem is how to address the issue with fabric-common not loading. So I am wondering why we have this type of definition would it not work with just ?
```
 packages: ['fabric-network@2.2.2', 'fabric-ca-client@2.2.2']
```
And if that's ok we apply it to all the 2.2.x definitions. If not then we probably need to add long to it
```
 packages: ['long@^4.0.0', 'fabric-common@2.2.2', 'fabric-protos@2.2.2', 'fabric-network@2.2.2', 'fabric-ca-client@2.2.2']
```

The other problem is the code that does the require, which is currently this logic
```
static moduleIsInstalled(moduleName, requireFunction = require) {
        let modulePath;
        if (moduleName.startsWith('./') || moduleName.startsWith('/')) {
            // treat it as an external module, but resolve the path, so it's absolute
            modulePath = CaliperUtils.resolvePath(moduleName);
        } else {
            // treat it as a package dependency (user must install it beforehand)
            modulePath = moduleName;
        }

        try {
            CaliperUtils.loadModule(modulePath, requireFunction);
            return true;
        } catch (err) {
            // If the module can not be found, an error is thrown
            return false;
        }
    }
```
What happens is that the long module couldn't be loaded but this code hides that error completely so you can't tell when the fabric-connector runs that it failed to load the module long. Improvements here are required to ensure we can easily find out what the problem is

davidkel:
closes #988

### DIFF
--- a/packages/caliper-cli/lib/lib/config.yaml
+++ b/packages/caliper-cli/lib/lib/config.yaml
@@ -48,11 +48,11 @@ sut:
         2.1.0:
             packages: ['fabric-common@2.1.0', 'fabric-protos@2.1.0', 'fabric-network@2.1.0', 'fabric-ca-client@2.1.0']
         2.2.0:
-            packages: ['fabric-common@2.2.0', 'fabric-protos@2.2.0', 'fabric-network@2.2.0', 'fabric-ca-client@2.2.0']
+            packages: ['fabric-network@2.2.0', 'fabric-ca-client@2.2.0']
         2.2.1:
-            packages: ['fabric-common@2.2.1', 'fabric-protos@2.2.1', 'fabric-network@2.2.1', 'fabric-ca-client@2.2.1']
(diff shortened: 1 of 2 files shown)


aklenik (approved):
Makes sense! 👍 
(The plugin/module loading should be extracted to a dedicated class later since it's quite an important feature.)

PR #994, opened 2020-09-11
Remove redundant event code from v1 gateway

davidkel:
For a gateway implementation it cannot support event compatibility mode
because the old event hub implementation was removed in 1.4 node-sdk.

The gateway implementation should not be used against fabrics prior to
1.4 either that would still have the old peer implementation of the
eventhub as 1.4 SDKs only support 1.4 fabrics and all SUTs prior to 1.4
have been removed in caliper

This PR removes the redundant code as it will never be used


aklenik (approved):
LGTM 👍

PR #1010, opened 2020-09-22
Add support for isTLSEnabled

davidkel:


aklenik:
Just a thought: the `_searchForPropertyValues` functionality could be generalized using JSON object queries, for example, with [jsonpath-plus](https://www.npmjs.com/package/jsonpath-plus). Then you could query every `url` value, and check whether TLS is used consistently. Might be worth investigating if we have more/many queries for the connection profile. Maintaining one-line queries would also be cheaper than maintaining iteration strategies when the object structure changes.

nklincoln (approved):

PR #1014, opened 2020-09-25
Add support for V2 Wallets

davidkel:
This implements a wallet facade that wrappers node-sdk v2 wallets for the new fabric connector configuration


nklincoln (changes requested):
Beautiful code, only a few super nit-picky comments on the test file. I suspect this is because no lint checks are run against the test code in the build perhaps

nklincoln (approved):

PR #1020, opened 2020-09-29
Make options robust and add unit tests to prom monitors

nklincoln:
When experimenting, I found that the current option settings for the prom monitors is not particularly robust. 

I have modified the constructors to pull from sensible defaults if not present, and added unit tests to support the prometheus monitors


packages/caliper-core/lib/common/config/Config.js
```diff
@@ -54,8 +54,12 @@ const keys = {
         }
     },
     Monitor: {
-        DefaultInterval: 'caliper-monitor-default-interval',
-        PrometheusScrapePort: 'caliper-monitor-prometheus-scrape-port'
+        Default: {
+            Interval: 'caliper-monitor-default-interval'
```
davidkel, line 58: Seems a bit odd to have a default section if the whole thing is to provide default values

packages/caliper-core/lib/common/config/default.yaml
```diff
@@ -75,9 +75,13 @@ caliper:
     # Configurations related to caliper test monitors
     monitor:
         # Default update interval
-        defaultinterval: 10000
-        # Default scrape port for prometheus tx observer
-        prometheusscrapeport: 3000
+        default:
+            # Default update interval
```
davidkel, line 79: replicated comment

packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js
```diff
@@ -0,0 +1,226 @@
...
+            }
+        });
+
+        await prometheusPushTxObserver.deactivate();
+
+        // Because we have stubbed out _sendUpdate, the default labels are not applied here
+        prometheusPushTxObserver.counterTxSubmitted.hashMap.should.deep.equal({
+            '': { value: 0, labels: {} }
+        });
+        prometheusPushTxObserver.counterTxFinished.hashMap.should.deep.equal({});
```
aklenik, line 223: 1. It's a bit strange that the two counters have different empty states. Is there a reason for that?
2. This comment is for the implementation itself rather than for the test. The stats won't overlap, since upon activating a new round, the basic labels will change, resulting in a separate time series. The current behavior will reset the counters to 0 which will remain there, possibly missing the max value in the end if scraped with unfortunate timing. So maybe we shouldn't reset the counters after the rounds? The label change is enough to isolate the rounds completely.

nklincoln: I think some experimenting around here is required. I completely agree that we don't want to throw away data, but i'm curious about the implications for this once benchmarks have been completed. With repeat benchmarks using dockerised workers, would we result in a continuous data line for all rounds ever run?

nklincoln: looking at the prom-client tests, the difference is to be expected: https://github.com/siimon/prom-client/blob/master/test/counterTest.js

aklenik: > With repeat benchmarks using dockerised workers, would we result in a continuous data line for all rounds ever run?

Yes, but this is true for any service, not just Caliper. If you want multiple independent runs in the same prometheus instance, then you must add a label that differentiates between them. Otherwise, it continues where it left off (to allow for service restarts). In the typical use cases, you will probably tear down prometheus along with Caliper and Fabric.

aklenik (approved):
LGTM 👍

PR #1030, opened 2020-10-01
Added support for Besu private transactions

lucassaldanha:
Besu supports [private transactions](https://besu.hyperledger.org/en/latest/Concepts/Privacy/Private-Transactions/) sent to a list of participant nodes. The library [web3js-eea](https://www.npmjs.com/package/web3-eea) supports sending private transactions to Besu.

By expanding the network configuration DSL and adding a few changes to the Ethereum connection, we can support sending private transactions to Besu using Caliper.

## Checklist
 - [X]  Design of the change
 - [X]  How to prove that the change works
 - [X]  Automated tests that prove the fix keeps on working

## Issue/User story
As a user, I want to have performance metrics of Besu when using private transactions.

## Steps to Reproduce
N/A

## Existing issues
N/A

## Design of the change
The core of the fix was to update `ethereum-connector.js` to be able to send private transactions. The network config DSL was extended to support some Besu privacy configurations.

In the network configuration, the `ethereum` object now can have a `privacy` section, enumerating privacy groups in the network. Also, there is the option of specifying what contracts should be private (and to what privacy group - aka participants - they belong to).

In the Ethereum connection, there are two new methods exclusively used for private transactions: `deployPrivateContract` and `_sendSinglePrivateRequest`.

## Validation of the change
I added a new step in the Besu integration tests to check the basic flow of public vs private contracts. I have done some manual testing as well to ensure the changes work.

## Automated Tests
See _Validation of the change_

I'm not sure if we have specific Besu documentation for Caliper. If we do, we could probably add some info about using Besu with private transactions. Otherwise, the integration tests should be a good start for anyone wanting to use it.

### DIFF
--- a/packages/caliper-ethereum/lib/ethereum-connector.js
+++ b/packages/caliper-ethereum/lib/ethereum-connector.js
@@ -16,6 +16,7 @@
 
 const EthereumHDKey = require('ethereumjs-wallet/hdkey');
 const Web3 = require('web3');
+const EEAClient = require('web3-eea');
 const {ConnectorBase, CaliperUtils, ConfigUtil, TxStatus} = require('@hyperledger/caliper-core');
 
 const logger = CaliperUtils.getLogger('ethereum-connector');
@@ -50,6 +51,9 @@ class EthereumConnector extends ConnectorBase {
 
         this.ethereumConfig = ethereumConfig;
         this.web3 = new Web3(this.ethereumConfig.url);
+        if (this.ethereumConfig.privacy) {
+            this.web3eea = new EEAClient(this.web3, ethereumConfig.chainId);
+        }
         this.web3.transactionConfirmationBlocks = this.ethereumConfig.transactionConfirmationBlocks;
         this.workerIndex = workerIndex;
         this.context = undefined;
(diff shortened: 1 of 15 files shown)


packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -100,13 +104,29 @@ class EthereumConnector extends ConnectorBase {
         let self = this;
         logger.info('Creating contracts...');
         for (const key of Object.keys(this.ethereumConfig.contracts)) {
-            let contractData = require(CaliperUtils.resolvePath(this.ethereumConfig.contracts[key].path)); // TODO remove path property
-            let contractGas = this.ethereumConfig.contracts[key].gas;
-            let estimateGas = this.ethereumConfig.contracts[key].estimateGas;
+            let contract = this.ethereumConfig.contracts[key];
+            let contractData = require(CaliperUtils.resolvePath(contract.path)); // TODO remove path property
+            let contractGas = contract.gas;
+            let estimateGas = contract.estimateGas;
```
benjamincburns, line 110: These could be `const` - helps the JIT optimizer

benjamincburns (approved):
LGTM - only one comment that's really inconsequential. Thanks a ton for including the new tests!

nklincoln:
@lucassaldanha  - there's a few LGTM bot warnings present at the moment, which you can view above, and one conflict due to a recent release that will need a quick rebase 👍

nklincoln:
@lucassaldanha - thanks for the update, there is one pending LGTM issue relating to the `store.js` test file in the constructor (this.contract has no effect) are you able to update that one too please?

lucassaldanha:
> thanks for the update, there is one pending LGTM issue relating to the `store.js` test file in the constructor (this.contract has no effect) are you able to update that one too please?

@nklincoln I thought I had it fixed might've missed it. Where are you seeing the warning?

nklincoln:
@lucassaldanha - thanks, I must have been looking at an out of date notification.

I've been looking at the output of the integration test run, and there appears to be a negative send rate being reported. Would you be able to bump the transaction number to a higher number (100 perhaps) so that we can determine if it is something in the private transaction addition, or a bug in caliper for low transaction numbers?

lucassaldanha:
@nklincoln Done! :)

packages/caliper-tests-integration/besu_tests/store.js
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        this.nodeUrl = sutContext.url;
+
+        if(this.roundArguments.private) {
+            this.isPrivate = true;
+            this.privacyOpts = sutContext.privacy[this.roundArguments.private];
+            this.privacyOpts['id'] = this.roundArguments.private;
+        } else {
+            this.private = false;
```
nklincoln, line 88: this.isPrivate = false ?

nklincoln (changes requested):
HI, had a deeper look as was seeing a negative send rate ... looks to be a bug in caliper main ... but did notice the flag that might need a change

packages/caliper-tests-integration/besu_tests/store.js
lucassaldanha: Nice catch! Because of the way the value was being consumed, it was undefined for public txs and in the end, the logic was working out the same. Lucky! :)

lucassaldanha: Fixed 👍

nklincoln:
@lucassaldanha  - Thanks for the contribution 👍 

It would be great if you could extend the docs to include details on how to take advantage of the new features added.

If you checkout the gh-pages branch, the file of interest is doc/vNext/EthereumConfiguration.md - any questions, please reach out to us on RC 😄

nklincoln (approved):

PR #1041, opened 2020-10-06
Prevent failures in integration tests

nklincoln:

Travis is not permitting access to create folders, so the ethereum tests are showing invalid errors. 

This PR modifies the ethereum/besu integration tests (report path)  to prevent the error


aklenik:
@nklincoln Any reason for not generating the report in the existing workspace dir (`./report.html`)? Not that it matters, just looks strange :)

nklincoln:
Was replicating the config from the Fabric connector tests if i'm honest! Happy to move all locations in this PR?

aklenik:
If we can't test the dir creation in Travis, then I think the default path is enough, so one less config to maintain 😁

aklenik (approved):

PR #1042, opened 2020-10-07
New V1 Gateway Connector

davidkel:
Test only new V1 Gateway Connector


packages/caliper-fabric/test/connector-versions/v1/FabricGateway.js
```diff
@@ -0,0 +1,529 @@
...
+                    invokerIdentity: 'User1',
+                    targetPeers: ['peer1', 'peer2']
+                };
+                await fabricGateway._sendSingleRequest(request);
+                Gateway.channel.should.equal('your-channel');
+                Transaction.submit.should.be.true;
+                Transaction.submitArgs.should.deep.equal(['arg1']);
+                Transaction.constructorArgs.should.equal('myFunction');
+                should.equal(Transaction.endorsingPeers, undefined);
+
```
nklincoln, line 386: commented out to be reinstated later?

packages/caliper-fabric/lib/connector-versions/v1/FabricGateway.js
```diff
@@ -0,0 +1,441 @@
...
+ *           are specified then targetPeers will take precedence
+ */
+
+/////////////////////////////
+// END OF TYPE DEFINITIONS //
+/////////////////////////////
+
+/**
+ */
+class V1FabricGateway extends ConnectorBase {
```
nklincoln, line 71: Do we want this to be named the same as the exported class?

nklincoln (changes requested):
Only a few questions on this one 👍

packages/caliper-fabric/test/connector-versions/v1/FabricGateway.js
davidkel: Yep, thanks need to remove that

packages/caliper-fabric/lib/connector-versions/v1/FabricGateway.js
davidkel: I was debating this with myself but would then see the need to split the connectors into 2 directories redundant (we would then also change the wallet names as well). I couldn't decide either so let it as is to see what comments came back :-)

nklincoln (approved):
🥇

PR #1045, opened 2020-10-08
Update txCollector and add unit tests

### LINKED ISSUE #1044: Negative send rate can be observed in reports
In some instances it is possible to observe a negative send rate in the reports.

This appears to occur in short round durations, and a quick set of logging in the report generation side (within getResultValues) reveals that there may be any issue in the txUpdate phase, since:

```
results.getFirstCreateTime() 1602161955354
results.getLastCreateTime()  1602161954519
```

At no point should the last create time be less than the first create time.

nklincoln:

closes #1044 

Fixes bug and add unit tests

### DIFF
--- a/packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
+++ b/packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
@@ -1,0 +1,356 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const TransactionStatisticsCollector = require('../../../lib/common/core/transaction-statistics-collector');
+const TxStatus = require('../../../lib/common/core/transaction-status');
+
+const chai = require('chai');
+chai.should();
+const sinon = require('sinon');
+
+describe('TransactionStatisticsCollector implementation', () => {
+
+    let sandbox;
+    let clock;
+
+    beforeEach(function() {
+        sandbox = sinon.createSandbox();
+        clock = sandbox.useFakeTimers();
+    });
+
+    afterEach(function() {
+        sandbox.restore();
+    });
+
+    describe('Using a single TransactionStatisticsCollector to collect results', () => {
+
+        it('should be possible to get basic round information from the collector', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector(0, 1, 'roundLabel');
+
+            transactionStatisticsCollector.getWorkerIndex().should.equal(0);
+            transactionStatisticsCollector.getRoundIndex().should.equal(1);
+            transactionStatisticsCollector.getRoundLabel().should.equal('roundLabel');
+        });
+
+        it('should not be possible to update a TransactionStatisticsCollector submission count if it is not activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+
+            transactionStatisticsCollector.txSubmitted(1);
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(0);
+        });
+
+        it('should not be possible to update a TransactionStatisticsCollector finished count if it is not activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished(txStatus);
+            sinon.assert.notCalled(updateSpy);
+        });
+
+        it('should be possible to update a TransactionStatisticsCollector with a single submission count once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            transactionStatisticsCollector.txSubmitted(1);
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(1);
+        });
+
+        it('should be possible to update a TransactionStatisticsCollector with a multiple submission counts once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            transactionStatisticsCollector.txSubmitted(7);
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(7);
+        });
+
+        it('should be possible to update a TransactionStatisticsCollector with a TxStatus result once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            // Must increment time for it to be valid
+            clock.tick(50);
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished(txStatus);
+            sinon.assert.calledOnce(updateSpy);
+        });
+
+        it('should be possible to update a TransactionStatisticsCollector with an array of TxStatus results once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            // Must increment time for it to be valid
+            clock.tick(50);
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished([txStatus, txStatus]);
+            sinon.assert.calledTwice(updateSpy);
+        });
+
+        it('should not be possible to update a TransactionStatisticsCollector with a transaction submission once deactivated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            transactionStatisticsCollector.txSubmitted(1);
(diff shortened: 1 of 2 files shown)


packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+        clock = sandbox.useFakeTimers();
+    });
+
+    afterEach(function() {
+        sandbox.restore();
+    });
+
+    describe('Using a single TransactionStatisticsCollector to collect results', () => {
+
+        it('should not be possible to get basic round information from the collector', () => {
```
aklenik, line 40: I think it `should be possible` to do this.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+        it('should not be possible to update a TransactionStatisticsCollector if it is not activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            transactionStatisticsCollector.txSubmitted(1);
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(0);
+
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished(txStatus);
```
aklenik, line 57: Two different API methods (`txSubmitted` and `txFinished`) are tested together. Might be better to split the test into two.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+        it('should be possible to update a TransactionStatisticsCollector with a single TxStatus results once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            transactionStatisticsCollector.txSubmitted(1);
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(1);
+
+            clock.tick(50);
```
aklenik, line 70: Might be worth a quick comment, that advancing the time is needed so the collector deems the result relevant.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished(txStatus);
```
aklenik, line 72: Again, two methods tested in one test.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            sinon.assert.calledOnce(updateSpy);
+        });
+
+        it('should be possible to update a TransactionStatisticsCollector with an array of TxStatus results once activated', () => {
+            const transactionStatisticsCollector = new TransactionStatisticsCollector();
+            transactionStatisticsCollector.activate();
+
+            const updateSpy = sandbox.spy(transactionStatisticsCollector, '_updateStatistics');
+
+            transactionStatisticsCollector.txSubmitted(1);
```
aklenik, line 82: Do we need the submit part to test the finished part? (If yes, then 2 TXs should be submitted here, for consistency.)

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            transactionStatisticsCollector.getTotalSubmittedTx().should.equal(1);
+
+            clock.tick(50);
+            const txStatus = new TxStatus();
+            transactionStatisticsCollector.txFinished([txStatus, txStatus]);
+            sinon.assert.calledTwice(updateSpy);
+        });
+
+        it('should not be possible to update a TransactionStatisticsCollector with a TxStatus result once deactivated', () => {
```
aklenik, line 91: The active state is already tested above. So only the deactivated state should be tested, and separately for submitted and finished TXs.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+
+            transactionStatisticsCollector.getTotalFinishedTx().should.equal(1);
+            transactionStatisticsCollector.getTotalSuccessfulTx().should.equal(1);
+            transactionStatisticsCollector.getFirstCreateTime().should.equal(50);
+            transactionStatisticsCollector.getLastCreateTime().should.equal(50);
+            transactionStatisticsCollector.getFirstFinishTime().should.equal(100);
+            transactionStatisticsCollector.getLastFinishTime().should.equal(100);
+            transactionStatisticsCollector.getMinLatencyForSuccessful().should.equal(50);
+            transactionStatisticsCollector.getMaxLatencyForSuccessful().should.equal(50);
+            transactionStatisticsCollector.getTotalLatencyForSuccessful().should.equal(50);
```
aklenik, line 134: Maybe we should also assert that the corresponding "failed" metrics are not updated.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+
+            transactionStatisticsCollector.getTotalFinishedTx().should.equal(1);
+            transactionStatisticsCollector.getTotalFailedTx().should.equal(1);
+            transactionStatisticsCollector.getFirstCreateTime().should.equal(50);
+            transactionStatisticsCollector.getLastCreateTime().should.equal(50);
+            transactionStatisticsCollector.getFirstFinishTime().should.equal(100);
+            transactionStatisticsCollector.getLastFinishTime().should.equal(100);
+            transactionStatisticsCollector.getMinLatencyForFailed().should.equal(50);
+            transactionStatisticsCollector.getMaxLatencyForFailed().should.equal(50);
+            transactionStatisticsCollector.getTotalLatencyForFailed().should.equal(50);
```
aklenik, line 156: Maybe we should also assert that the corresponding "successful" metrics are not updated.

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            clock.tick(50);
+            const txStatus = new TxStatus();
+            clock.tick(50);
+            txStatus.SetStatusSuccess();
+
+            transactionStatisticsCollector.txFinished(txStatus);
+
+            transactionStatisticsCollector.getTotalFinishedTx().should.equal(1);
+            transactionStatisticsCollector.getTotalSuccessfulTx().should.equal(1);
+            transactionStatisticsCollector.getFirstCreateTime().should.equal(50);
```
aklenik, line 128: It might be better to use the `txStatus.GetXY()` return values for the expected values, rather than literal numbers. That will clearly indicate the source of the expected value. (Only applies if a metric is derived from txStat metrics.)

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            clock.tick(50);
+            const txStatus = new TxStatus();
+            clock.tick(50);
+            txStatus.SetStatusFail();
+
+            transactionStatisticsCollector.txFinished(txStatus);
+
+            transactionStatisticsCollector.getTotalFinishedTx().should.equal(1);
+            transactionStatisticsCollector.getTotalFailedTx().should.equal(1);
+            transactionStatisticsCollector.getFirstCreateTime().should.equal(50);
```
aklenik, line 150: It might be better to use the `txStatus.GetXY()` return values for the expected values, rather than literal numbers. That will clearly indicate the source of the expected value. (Only applies if a metric is derived from txStat metrics.)

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            const txStatus0 = new TxStatus();
+            clock.tick(25);
+            const txStatus1 = new TxStatus();
+            txStatus0.SetStatusSuccess();
+            clock.tick(100);
+            txStatus1.SetStatusSuccess();
+
+            transactionStatisticsCollector.txFinished([txStatus0, txStatus1]);
+            transactionStatisticsCollector.getTotalSuccessfulTx().should.equal(2);
+            transactionStatisticsCollector.getFirstCreateTime().should.equal(50);
```
aklenik, line 174: It might be better to use the `txStatus.GetXY()` return values for the expected values, rather than literal numbers. That will clearly indicate the source of the expected value. (Only applies if a metric is derived from txStat metrics.)

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
```diff
@@ -0,0 +1,219 @@
...
+            clock.tick(100);
+            txStatus1.SetStatusSuccess();
+
+            transactionStatisticsCollector0.txFinished(txStatus0);
+            transactionStatisticsCollector1.txFinished(txStatus1);
+
+            const mergedCollector = TransactionStatisticsCollector.mergeCollectorResults([transactionStatisticsCollector0, transactionStatisticsCollector1]);
+
+            mergedCollector.getTotalSuccessfulTx().should.equal(2);
+            mergedCollector.getFirstCreateTime().should.equal(50);
```
aklenik, line 208: It might be better to use the `txStatus.GetXY()` return values for the expected values, rather than literal numbers. That will clearly indicate the source of the expected value. (Only applies if a metric is derived from txStat metrics.)

aklenik (changes requested):
Just some minor clarity-aiding refactoring for the tests, otherwise nice fix 👍

packages/caliper-core/test/common/core/transaction-statisitcs-collector.js
nklincoln: yeah, otherwise whats the point 😂  ... best edit that!

aklenik (approved):
Awesome :+1:

PR #1050, opened 2020-10-12
Update node version in image

### LINKED ISSUE #1049: Caliper container failing
If using the caliper image to benchmark, it can fail and result in "hanging" the benchmark execution.

Scenario:
- Follow the tutorial, but apply testing with the caliper image
- Increase the loading applied during the benchmark round

## Observations
Worker logging reveals:
```
2020.10.09-16:42:47.038 info  [caliper] [caliper-worker]        Worker #1 starting workload loop
node[1]: ../src/node_http2.cc:893:ssize_t node::http2::Http2Session::ConsumeHTTP2Data(): Assertion `(flags_ & SESSION_STATE_READING_STOPPED) != (0)' failed.
```
After which the worker is essentially dead (if launched by manager, the manager process will continue, resulting in a continued null update from any txMonitors)

Is visible for all possible container operational modes:
- container with internal workers spawned
- container with internal workers using mqtt communication
- distributed workers using mqtt with a central manager (in this case the workers are the only items to report the error)

This appears to occur under high loading

There are issues raised against nodeJS that are in this area:
- https://github.com/nodejs/node/issues/29393

With indication that fixes have been added to 13.3.0, 12.14.1 and 10.18.1

Initial proposal is to jump immediately to 10.18 in the container (remaining within the v10 stream) and investigate v12 levels going forwards as a part of #965

nklincoln:
closes #1049 

Move to v10.22 node in caliper image 

### DIFF
--- a/packages/caliper-publish/caliper.Dockerfile
+++ b/packages/caliper-publish/caliper.Dockerfile
@@ -12,7 +12,7 @@
 # limitations under the License.
 #
 
-FROM node:10.16-alpine
+FROM node:10.22-alpine
 
 # require to set these explicitly to avoid mistakes
 ARG npm_registry


davidkel (approved):
LGTM

PR #1055, opened 2020-10-15
Extract identities from a version specific wallet and store in the in memory wallet

RosieMurphy0:
Implemented the _extractIdentitiesFromWallet method in the identity manager and the associated tests for the method. Takes identities from a wallet and stores them in the in memory wallet using an existing method.

contributes to #940

### DIFF
--- a/packages/caliper-fabric/lib/identity-management/IdentityManager.js
+++ b/packages/caliper-fabric/lib/identity-management/IdentityManager.js
@@ -191,7 +191,14 @@ class IdentityManager {
      * @private
      */
     async _extractIdentitiesFromWallet(mspId, wallet) {
-        // TODO: To be implemented
+        const walletFacade = await this.walletFacadeFactory.create(wallet.path);
+        const allIDNames = await walletFacade.getAllIdentityNames();
+        // eslint-disable-next-line no-console
+        console.log({allIDNames});
+        for (const identityName in allIDNames){
+            const identity = await walletFacade.export(identityName);
+            await this._addToWallet(identity.mspid, allIDNames[identityName], identity.certificate, identity.privateKey);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/caliper-fabric/lib/identity-management/IdentityManager.js
```diff
@@ -191,7 +191,14 @@ class IdentityManager {
      * @private
      */
     async _extractIdentitiesFromWallet(mspId, wallet) {
-        // TODO: To be implemented
+        const walletFacade = await this.walletFacadeFactory.create(wallet.path);
+        const allIDNames = await walletFacade.getAllIdentityNames();
```
davidkel, line 195: could you change `allIDNames` to `allIdentityNames` as we don't want to use abbreviations

packages/caliper-fabric/lib/identity-management/IdentityManager.js
```diff
@@ -191,7 +191,14 @@ class IdentityManager {
...
+        // eslint-disable-next-line no-console
+        console.log({allIDNames});
```
davidkel, line 197: Looks like you left some debug code in, could you delete the console.log and the comment above it as well.

packages/caliper-fabric/lib/identity-management/IdentityManager.js
```diff
@@ -191,7 +191,14 @@ class IdentityManager {
...
+        for (const identityName in allIDNames){
```
davidkel, line 198: allIDNames to allIdentityNames here too

packages/caliper-fabric/lib/identity-management/IdentityManager.js
```diff
@@ -191,7 +191,14 @@ class IdentityManager {
...
+            const identity = await walletFacade.export(identityName);
+            await this._addToWallet(identity.mspid, allIDNames[identityName], identity.certificate, identity.privateKey);
```
davidkel, line 200: with the change to `of` in the for loop, you can replace the `allIDNames[identityName]` with just `identityName`

PR #1058, opened 2020-10-16
Extract identities from wallet to memory wallet

RosieMurphy0:
Implemented the _extractIdentitiesFromWallet method in the identity manager and the associated tests for the method. Takes identities from a wallet and stores them in the in memory wallet using an existing method. Changed some tests in the connector file as well that were effected. 

contributes to #940


davidkel (approved):
LGTM

PR #1063, opened 2020-10-19
update version to unstable

nklincoln:
bump versions to 0.4.2-unstable


aklenik (changes requested):
The new version should be `0.4.2-unstable`.

aklenik (approved):

PR #1089, opened 2020-11-17
support multiorg and channels

davidkel:
Need to ensure we don't stop working when precaching the environment
because in a multiorg definition they may not have access to all the
channels defined.
This is an edge case because multiple org definitions is not a good use
of this implementation. But because it supports it we want to remove
this restriction.


nklincoln (approved):
👍

PR #1105, opened 2020-12-23
Azure pipelines build

nklincoln:
Contributes to #1102 

Introduces Azure pipeline build
- unit tests
- integration tests


.travis.yml
```diff
@@ -10,7 +10,7 @@ dist: xenial
 before_install: |
     set -ev
     set -o pipefail
-    npm install -g @alrra/travis-scripts
+    npm install -g @alrra/build-scripts
```
aklenik, line 13: Is this rename intentional?

azure-pipelines.yml
```diff
@@ -0,0 +1,96 @@
+# Node.js
+# Build a general Node.js project with npm.
+# Add steps that analyze code, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
```
aklenik, line 4: Do we need license in yaml files?

aklenik (approved):
Two non-blocking comments only

.travis.yml
nklincoln: 🤔  don't actually remember changing this on, i will revert the change

azure-pipelines.yml
nklincoln: Looks to be a bit of self publication from microsoft - it was an autogenerated document

PR #1135, opened 2021-04-23
Resolve issue in which only first round would be docker-monitored

### LINKED ISSUE #1134: Docker Monitor with multiple rounds only monitors first round
When a benchmark is configured with a docker monitor and multiple rounds, only the results of the first round are printed and attached to the report.

## Context
1. Monitors are defined on the root level, i.e. for every round. Besides, process monitors do monitor all rounds, instead of just the first one.

## Expected Behavior
Each round should be monitored and exported.

## Actual Behavior
Only first round is monitored and exported.

## Possible Fix
Modify `if (!this.containers) {` to `if (!this.containers || this.containers.length == 0) {` in `monitor-docker.js` line 277.

## Steps to Reproduce
1. Start any docker monitor on a multi-round benchmark, i.e. the besu phase3 benchmark in the integration tests with added docker monitor.

## Existing issues
N/A

## Context
Should be evident.

## Your Environment
v0.4.2 as well as the master branch version.

Viserius:
There is an issue that docker containers are only monitored during the first round of a multi-round benchmark as described in issue #1134.
The problem is caused by the fact that the `stop()` command sets the `this.containers` property to the empty array `[]`, while it expects a `null` value to (re)set the containers list. By setting this to `if (!this.containers || this.containers.length == 0)`, we account for both cases.

Resolves #1134

### DIFF
--- a/packages/caliper-core/lib/manager/monitors/monitor-docker.js
+++ b/packages/caliper-core/lib/manager/monitors/monitor-docker.js
@@ -274,7 +274,7 @@ class MonitorDocker extends MonitorInterface {
      */
     async start() {
         // Conditionally build monitored containers, these are persisted between rounds and restart action
-        if (!this.containers || this.containers.length == 0) {
+        if (!this.containers || this.containers.length === 0) {
             await this.findContainers();
         }
         // Read stats immediately, then kick off monitor refresh at interval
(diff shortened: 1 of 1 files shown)


nklincoln:
thanks for the contribution - you will need to sign off the commit before we can merge (DCO bot rules) and once we can figure out the issue in the pipeline with the Ethereum adaptor, we can merge it in

Viserius:
Thank you for highlighting this, I had not seen that!

nklincoln (approved):

nklincoln:
@aklenik - Seems the azure re-run builds is not working at the moment - i'm happy for this to go in "as is" and with the ethereum failure based on the fact that the other integration tests would have used this code change

PR #1138, opened 2021-04-27
Allow specification of custom EOA's and nonces

Viserius:
Proposes a solution to #1135 , namely by allowing users to set request.sender.privateKey and request.sender.nonce.

### DIFF
--- a/packages/caliper-ethereum/lib/ethereum-connector.js
+++ b/packages/caliper-ethereum/lib/ethereum-connector.js
@@ -220,7 +220,7 @@ class EthereumConnector extends ConnectorBase {
 
         const context = this.context;
         let status = new TxStatus();
-        let params = {from: context.fromAddress};
+        let params = {};
         if (request.hasOwnProperty('value') && request.value > 0) {
             params.value = request.value;
         }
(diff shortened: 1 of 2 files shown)


nklincoln:
similar to you other PR, would you be able to sign off the commits so that once the pipeline issue is resolved, we can merge? Thanks!

Viserius:
Bump

aklenik:
Closing and reopening to trigger updated CI checks

packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -255,32 +251,41 @@ class EthereumConnector extends ConnectorBase {
...
-            } catch (err) {
-                onFailure(err);
-            }
-        } else {
-            if (contractInfo.gas && contractInfo.gas[request.verb]) {
-                params.gas = contractInfo.gas[request.verb];
-            } else if (contractInfo.estimateGas) {
-                params.gas = 1000 + await contractInfo.contract.methods[request.verb].estimateGas(params);
-            }
+                if (request.sender) {
```
aklenik, line 262: Shouldn't this block also be present in the `!request.args` path? Because right now the custom sender can only be used with transactions that require some arguments. Or am I missing something?

aklenik (commented):
@Viserius Could you clarify some things, please? (And sorry for the late review)

packages/caliper-ethereum/lib/ethereum-connector.js
Viserius: Hello @aklenik, it has been ages since I worked with this. Looking back at my project, you are probably right.

You can check in the following file how I used this technique (round-robin): https://github.com/Viserius/FabricBesuBenchmark/blob/master/caliper-workspace/workload/besunft-transfer.js

As you can see, I had it put in a local args variable, which might have caused the confusion of putting it in the request.args branch. Would you mind making the necessary adjustments?

aklenik: Thanks for the quick response! I'll think about the correct branching strategy, and push a new commit to your branch (and will ask @davidkel to review the PR, since I probably can't review my own contribution 😄)

PR #1164, opened 2021-09-21
Added basic value providers in workload module

bistaastha:
- Added value-provider-interface and uniform-random-value-provider
- Added unit tests for value providers
- UniformRandomValueProvider test fix and reference value provider addition

Signed-off-by: Aastha Bist <abist119@gmail.com>

### DIFF
--- a/packages/caliper-core/lib/worker/workload/declarative/value-providers/parameter-reference-value-provider.js
+++ b/packages/caliper-core/lib/worker/workload/declarative/value-providers/parameter-reference-value-provider.js
@@ -1,0 +1,63 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const ValueProviderInterface = require('./value-provider-interface');
+
+/**
+ * Class with implementation of ParameterReferenceValueProvider
+ */
+class ParameterReferenceValueProvider extends ValueProviderInterface {
+    /**
+     * Initialize an instance of ParameterReferenceValueProvider
+     * @param {object} options Options for configured parameter generation
+     * @param {object} variables Caliper provided variables
+     * @param {object} parameters User provided variables
+     * @param {object} valueProviderFactory ValueProviderFactory object reference
+     */
+    constructor(options, variables, parameters, valueProviderFactory) {
+        super(options, variables, parameters, valueProviderFactory);
+
+        if(this.options === undefined) {
+            throw new Error(`Incorrect options value: ${this.options}`);
+        }
+
+        if(this.parameters === undefined) {
+            throw new Error(`Incorrect parameters value: ${this.parameters}`);
+        }
+
+        if (
+            this.options.name === undefined ||
+            typeof this.options.name !== 'string'
+        ) {
+            throw new Error(
+                `Incorrect parameter name: ${options.name}`
+            );
+        }
+
+        if (this.parameters[this.options.name] === undefined) {
+            throw new Error(`Parameter ${this.options.name} cannot be found among available parameters`);
+        }
+    }
+    /**
+     * Generates value for corresponding parameter
+     * @returns {any} value for parameter according to options.name
+     */
+    generateValue() {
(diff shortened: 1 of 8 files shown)


packages/caliper-core/lib/worker/workload/declarative/value-providers/value-provider-interface.js
```diff
@@ -0,0 +1,49 @@
...
+class ValueProviderInterface {
+    /**
+     * Initialize an instance of ValueProviderInterface
+     * @param {object} options Options for configured parameter generation
+     * @param {object} variables Caliper provided variables
+     * @param {object} parameters User provided variables
+     * @param {object} valueProviderFactory ValueProviderFactory object reference
+     */
+    constructor(
+        options = {},
```
aklenik, line 28: On second thought, we shouldn't assign default values to these parameters. If something is not passed, it should be undefined, and checked later by inheriting classes.

packages/caliper-core/lib/worker/workload/declarative/value-providers/value-provider-interface.js
```diff
@@ -0,0 +1,49 @@
...
+        variables = {},
+        parameters = {},
+        valueProviderFactory
+    ) {
+        this.options = options;
+        this.variables = variables;
+        this.parameters = parameters;
+        if (valueProviderFactory === undefined) {
```
aklenik, line 36: This check should be left to inheriting classes, if they need the factory. We should remain flexible in the base class.

packages/caliper-core/lib/worker/workload/declarative/value-providers/value-provider-interface.js
```diff
@@ -0,0 +1,49 @@
...
+            throw new Error('valueProviderFactory is undefined');
+        }
+        this.valueProviderFactory = valueProviderFactory;
+    }
+    /**
+     * Generate values for corresponding options provided
+     */
```
aklenik, line 43: The doc should contain a return statement (with type `object`). And the code should also contain a `return undefined;` after the `throw` clause, just to pass lint checks.

packages/caliper-core/lib/worker/workload/declarative/value-providers/variable-reference-value-provider.js
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Initialize an instance of VariableReferenceValueProvider
+     * @param {object} options Options for configured parameter generation
+     * @param {object} variables Caliper provided variables
+     * @param {object} parameters User provided variables
+     * @param {object} valueProviderFactory ValueProviderFactory object reference
+     */
+    constructor(options, variables, parameters, valueProviderFactory) {
+        super(options, variables, parameters, valueProviderFactory);
+
```
aklenik, line 32: It should be checked whether `this.options` is undefined, and throw an error if it is.

packages/caliper-core/lib/worker/workload/declarative/value-providers/variable-reference-value-provider.js
```diff
@@ -0,0 +1,55 @@
...
+        if (
+            this.options.name === undefined ||
+            typeof this.options.name !== 'string'
+        ) {
+            throw new Error(
+                `Incorrect variable name: ${options.name}`
+            );
+        }
+
```
aklenik, line 41: It should be checked whether `this.variables` is undefined, and throw an error if it is.

packages/caliper-core/lib/worker/workload/declarative/value-providers/parameter-reference-value-provider.js
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Initialize an instance of ParameterReferenceValueProvider
+     * @param {object} options Options for configured parameter generation
+     * @param {object} variables Caliper provided variables
+     * @param {object} parameters User provided variables
+     * @param {object} valueProviderFactory ValueProviderFactory object reference
+     */
+    constructor(options, variables, parameters, valueProviderFactory) {
+        super(options, variables, parameters, valueProviderFactory);
+
```
aklenik, line 32: It should be checked whether `this.options` is undefined, and throw an error if it is.

aklenik (changes requested):
A few more checks are still needed once the default parameter values are changed (back to undefined). Tests should be updated accordingly.

packages/caliper-core/lib/worker/workload/declarative/value-providers/value-provider-interface.js
bistaastha: Doing this fails the checks as there is unreachable return undefined after the throw is unreachable code.

aklenik: @bistaastha I think you can disable that rule only for that line, with a remark stating the reason (e.g., documentation correctness is more important).

aklenik (approved):

PR #1168, opened 2021-10-06
Add factory class, formatted string and list element value providers.

bistaastha:
Signed-off-by: Aastha Bist <abist119@gmail.com>

### DIFF
--- a/packages/caliper-core/lib/worker/workload/declarative/value-providers/formatted-string-value-provider.js
+++ b/packages/caliper-core/lib/worker/workload/declarative/value-providers/formatted-string-value-provider.js
@@ -1,0 +1,99 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


packages/caliper-core/test/worker/workload/declarative/value-providers/value-provider-factory.test.js
```diff
@@ -0,0 +1,134 @@
...
+
+    describe('createValueProvider', () => {
+
+        it('should throw an error on undefined type', () => {
+            const wrapper = () => new ValueProviderFactory ({}, {}).createValueProvider(undefined, {});
+            expect(wrapper).to.throw(Error, 'Invalid value for type: undefined');
+        });
+
+        it('should return formatted string value provider object', () => {
+            const wrapper = () => new ValueProviderFactory ({}, {}).createValueProvider(undefined, {});
```
aklenik, line 35: Copy-paste error, should create a string value provider and check for it.

packages/caliper-core/lib/worker/workload/declarative/value-providers/formatted-string-value-provider.js
```diff
@@ -0,0 +1,108 @@
...
+
+        if (!Array.isArray(this.options.parts)) {
+            throw new Error(`Incorrect parts value: ${this.options.parts}`);
+        }
+
+        this.format = this.options.format;
+        this.parts = this.options.parts;
+        const partLength = this.parts.length;
+
+        for (let i = 1; i <= partLength; i++) {
```
aklenik, line 55: This check is not needed, since Line#78 does the same in a nicer way. Also update the corresponding expected error messages in the tests

aklenik (changes requested):
Two minor changes requested, then it's good to go :)

aklenik (approved):

PR #1174, opened 2021-11-15
Fix Besu integration test

aklenik:
* Change opened RPC WS API names to upper case, lower case names resulted in error.
* Remove the custom docker build from the network, use the Besu image directly and set the required parameters on the container.
* Fix the Besu version to 21.10.1
* Add some docker container log dumps in case of CI test failure


davidkel (approved):
LGTM

PR #1183, opened 2021-12-07
Fix issue where init flow errors when not using mutual TLS

### LINKED ISSUE #1160: New fabric connector crashes if you try to use the init flow without mutual TLS
The init flow in the new fabric 1.4 connector can configure a fabric network but should not be used outside of the caliper build environment. As it uses Mutual TLS it works. The fabric 2.2 init flow doesn't do anything except log. The problem is both of the init flows make the following call

```
        const tlsInfo = this.connectorConfiguration.isMutualTLS() ? 'mutual'
            : (this.connectorConfiguration.getConnectionProfileDefinitionForOrganization(defaultOrganization).isTLSEnabled() ? 'server' : 'none');
```

The issue here is that `getConnectionProfileDefinitionForOrganization` returns a promise, so it is missing an `await`.

We should fix this for 1.4, but for 2.2 given that init does nothing and there is no point using an init flow, we should no-op this in the 2.2 connector.

davidkel:
although init flows should not be used with the new connectors if you do it shouldn't fail but it fails with the following error if you aren't using mutual TLS

TypeError: this.connectorConfiguration.getConnectionProfileDefinitionForOrganization(...).isTLSEnabled is not a function

closes #1160


aklenik (approved):
LGTM 👍

Issue #1159, opened 2021-09-14
Caliper docker metrics captures incorrect memory usage


davidkel:
The docker stats api documentation which can be found here
https://docs.docker.com/engine/api/v1.41/#operation/ContainerStats

shows that memory usage is calculated as follows
```
used_memory = memory_stats.usage - memory_stats.stats.cache
```
However caliper doesn't do this which will result in wrong usage statistics, eg I've seen 13.8Gb and 21Gb being reported

PR #1214, opened 2022-01-28
Docker monitor metrics fixes - cpu and memory usage

### LINKED ISSUE #1166: Docker monitor should normalize CPU performances according to number of cores
Caliper should prouduce CPU performances on a scale of 100 (and not over).

Docker monitor (`packages/caliper-core/lib/manager/monitors/monitor-docker.js`) needs to consider the number of cores the machine disposes of.

Divide total CPU usage by number of cores => CPU usage/ cores

- [x] Add `/ os.cpus().length ` in function `getStatistics() ` when adding cpu utilization to the stats to retrieve: `watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2) /  os.cpus().length ); watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2) / os.cpus().length );`

- [x] integrate optional field on the `benchmark config file` #1214 
          - set defult to false to keep existing behaviour

- [x]  add documentation #1215 

- [x] add related unit test

```
               monitors:
                  resource:
                  - module: docker
                    options:
                      cpuUsageNormalization: true     
```

fraVlaca:
fixes #1166 and #1159.

- adds normalization option for cpu usage metric of docker-monitor 
- fixes memory usage stat of docker-monitor

### DIFF
--- a/packages/caliper-core/lib/manager/monitors/monitor-docker.js
+++ b/packages/caliper-core/lib/manager/monitors/monitor-docker.js
@@ -20,7 +20,7 @@ const Logger = Util.getLogger('monitor-docker');
 const MonitorInterface = require('./monitor-interface');
 const MonitorUtilities = require('./monitor-utilities');
 const ChartBuilder = require('../charts/chart-builder');
-
+const os = require('os');
 const URL = require('url');
 const Docker = require('dockerode');
 const SystemInformation = require('systeminformation');
@@ -338,6 +338,7 @@ class MonitorDocker extends MonitorInterface {
 
         const resourceStats = [];
         let chartStats = [];
+        let cpuCores = 1;
         try {
             // Build a statistic for each monitored container and push into watchItems array
             for (const container of this.containers) {
@@ -359,8 +360,11 @@ class MonitorDocker extends MonitorInterface {
                     watchItemStat.set('Name', container.name);
                     watchItemStat.set('Memory(max)', mem_stat.max);
                     watchItemStat.set('Memory(avg)', mem_stat.avg);
-                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2));
-                    watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2));
+                    if(this.options.hasOwnProperty('cpuUsageNormalization')){
+                        cpuCores = os.cpus().length;
+                    }
+                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2))/cpuCores;
+                    watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2))/cpuCores;
                     watchItemStat.set('Traffic In', (net.in[net.in.length - 1] - net.in[0]));
                     watchItemStat.set('Traffic Out', (net.out[net.out.length - 1] - net.out[0]));
                     watchItemStat.set('Disc Write', (disc.write[disc.write.length - 1] - disc.write[0]));
--- a/packages/caliper-core/lib/manager/monitors/monitor-docker.js
+++ b/packages/caliper-core/lib/manager/monitors/monitor-docker.js
@@ -340,7 +340,6 @@ class MonitorDocker extends MonitorInterface {
 
         const resourceStats = [];
         let chartStats = [];
-        let cpuCores = 1;
         try {
             // Build a statistic for each monitored container and push into watchItems array
             for (const container of this.containers) {
@@ -362,13 +361,12 @@ class MonitorDocker extends MonitorInterface {
                     watchItemStat.set('Name', container.name);
                     watchItemStat.set('Memory(max)', mem_stat.max);
                     watchItemStat.set('Memory(avg)', mem_stat.avg);
-                    if(this.options.hasOwnProperty('cpuUsageNormalization')){
-                        if(this.options.cpuUsageNormalization){
-                            cpuCores = os.cpus().length;
-                        }
+                    let cpuCores = 1;
+                    if (this.options.cpuUsageNormalization) {
+                        cpuCores = os.cpus().length;
                     }
-                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2)/cpuCores);
-                    watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2)/cpuCores);
+                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2) / cpuCores);
+                    watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2) / cpuCores);
                     watchItemStat.set('Traffic In', (net.in[net.in.length - 1] - net.in[0]));
                     watchItemStat.set('Traffic Out', (net.out[net.out.length - 1] - net.out[0]));
                     watchItemStat.set('Disc Write', (disc.write[disc.write.length - 1] - disc.write[0]));
--- a/packages/caliper-core/lib/manager/monitors/monitor-docker.js
+++ b/packages/caliper-core/lib/manager/monitors/monitor-docker.js
@@ -201,8 +201,9 @@ class MonitorDocker extends MonitorInterface {
                     }
                     if (this.containers[i].remote === null) {
                         // local
-                        this.stats[id].mem_usage.push(stat.mem_usage);
-                        this.stats[id].mem_percent.push(stat.mem_percent);
+                        const actualMemUsage = stat.memory_stats.usage - stat.memory_stats.stats.cache;
+                        this.stats[id].mem_usage.push(actualMemUsage);
+                        this.stats[id].mem_percent.push(actualMemUsage / stat.mem_limit);
                         let cpuDelta = stat.cpu_stats.cpu_usage.total_usage - stat.precpu_stats.cpu_usage.total_usage;
                         let sysDelta = stat.cpu_stats.system_cpu_usage - stat.precpu_stats.system_cpu_usage;
                         if (cpuDelta > 0 && sysDelta > 0) {
@@ -220,8 +221,9 @@ class MonitorDocker extends MonitorInterface {
                         this.stats[id].blockIO_wx.push(stat.blockIO.w);
                     } else {
                         // remote
-                        this.stats[id].mem_usage.push(stat.memory_stats.usage);
-                        this.stats[id].mem_percent.push(stat.memory_stats.usage / stat.memory_stats.limit);
+                        const actualMemUsage = stat.memory_stats.usage - stat.memory_stats.stats.cache;
+                        this.stats[id].mem_usage.push(actualMemUsage);
+                        this.stats[id].mem_percent.push(actualMemUsage / stat.mem_limit);
                         //this.stats[id].cpu_percent.push((stat.cpu_stats.cpu_usage.total_usage - stat.precpu_stats.cpu_usage.total_usage) / (stat.cpu_stats.system_cpu_usage - stat.precpu_stats.system_cpu_usage) * 100);
                         let cpuDelta = stat.cpu_stats.cpu_usage.total_usage - stat.precpu_stats.cpu_usage.total_usage;
                         let sysDelta = stat.cpu_stats.system_cpu_usage - stat.precpu_stats.system_cpu_usage;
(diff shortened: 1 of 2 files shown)


davidkel (changes requested):
Please could you include a unit test with this PR

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+                    name:"peer0.org1.example.com",
+                    remote:null
+                }
+            ]
+            await dockerMonitor.findContainers().then(()=>{
+                dockerMonitor.containers.should.be.an('array').that.deep.equals(containers);
+            });
+        });
+
+        /*it('should add and label containers as remote', async () => {
```
davidkel, line 131: We should remove commted out code

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+            }).then(async () => {
+                return await dockerMonitor.getStatistics()
+            }).then(function(result) {
+                result.resourceStats[0].get("CPU%(max)").should.equal(cpu_percent/cpuCores);
+            });
+        });
+    });
+
+
+    describe.skip('when stopping', () => {
```
davidkel, line 204: this should be removed

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+            if (dockerStats.cpu_stats.cpu_usage.hasOwnProperty('percpu_usage') && dockerStats.cpu_stats.cpu_usage.percpu_usage !== null) {
+                cpu_percent = (cpuDelta / sysDelta * cpuCores * 100.0);
+            } else {
+                cpu_percent = (cpuDelta / sysDelta * 100.0);
+            }
+        } else {
+            cpu_percent = 0;
+        }
+
+        it('should normalize cpuUsage if cpuUsageNormalization property is present', async () => {
```
davidkel, line 191: There should be 2 more tests
1. should not normalize if property set to false
2. should not normalize if property not provided

test should say `should normalize cpu usage if cpuUsageNormalization property is set to true`

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+            await dockerMonitor.findContainers().then(async () => {
+                await dockerMonitor.readContainerStats()
+            }).then(async () => {
+                dockerMonitor.stats["118"].mem_usage.should.be.an('array').that.deep.equals([mem_usage]);
+            });
+        });
+
+    });
+
+    describe('when getting statistics', () => {
```
davidkel, line 169: there are 2 describes which basically say the same thing, there needs to be just a single describe of `when retrieving container statistics` with all the tests under it

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+        cpu_percent: 10000,
+        netIO: {
+          rx: 10000,
+          wx: 10000,
+        },
+        blockIO: {
+          r: 10000,
+          w: 10000,
+        },
+        cpu_stats: {"cpu_usage":{"total_usage":1000000,"usage_in_kernelmode":10000,"usage_in_usermode":10000},"system_cpu_usage":100000,"online_cpus":8,"throttling_data":{"periods":0,"throttled_periods":0,"throttled_time":0}},
```
davidkel, line 45: Is it possible to format this json a bit more so that it's easier to read

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+        id:"118",
+        name:"peer0.org1.example.com"
+    }]);
+
+    SystemInformationStub.dockerContainerStats.withArgs("118").returns(dockerStats);
+
+    DockerMonitorRewire.__set__("SystemInformation", SystemInformationStub)
+    let dockerMonitor;
+
+    before(() => {
```
davidkel, line 67: As there is no implementation in these sections they don't need to be declared

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+    
+    SystemInformationStub.dockerContainers.returns([{
+        id:"118",
+        name:"peer0.org1.example.com"
+    }]);
+
+    SystemInformationStub.dockerContainerStats.withArgs("118").returns(dockerStats);
+
+    DockerMonitorRewire.__set__("SystemInformation", SystemInformationStub)
+    let dockerMonitor;
```
davidkel, line 65: There is no need for dockerMonitor to be declared here as there would be no need to share it across anything. Each of the tests can just do `let dockerMonitor = ...`

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+    });
+
+    // Test data
+    const monitorOptions = {
+        interval: 5,
+        containers: ["peer0.org1.example.com"],
+        cpuUsageNormalization: true  
+    };
+
+    const emptyOptions = {};
```
davidkel, line 85: emptyOptions is only used by 1 test, so we can move this declaration into that test

packages/caliper-core/lib/manager/monitors/monitor-docker.js
```diff
@@ -338,6 +340,7 @@ class MonitorDocker extends MonitorInterface {
 
         const resourceStats = [];
         let chartStats = [];
+        let cpuCores = 1;
```
davidkel, line 343: we can declare this variable closer to where it is first used, see next comment

packages/caliper-core/lib/manager/monitors/monitor-docker.js
```diff
@@ -359,8 +362,13 @@ class MonitorDocker extends MonitorInterface {
                     watchItemStat.set('Name', container.name);
                     watchItemStat.set('Memory(max)', mem_stat.max);
                     watchItemStat.set('Memory(avg)', mem_stat.avg);
-                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2));
-                    watchItemStat.set('CPU%(avg)', cpu_stat.avg.toFixed(2));
+                    if(this.options.hasOwnProperty('cpuUsageNormalization')){
```
davidkel, line 365: We can make this simpler as follows
```
let cpuCores = 1;
if (this.options.cpuUsageNormalization) {
    cpuCores = os.cpus().length;
}
```

packages/caliper-core/lib/manager/monitors/monitor-docker.js
```diff
@@ -359,8 +362,13 @@ class MonitorDocker extends MonitorInterface {
...
+                        if(this.options.cpuUsageNormalization){
+                            cpuCores = os.cpus().length;
+                        }
+                    }
+                    watchItemStat.set('CPU%(max)', cpu_stat.max.toFixed(2)/cpuCores);
```
davidkel, line 370: for coding guidelines we like to have spaces surrounding the `/`

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,211 @@
...
+
+    });
+
+    describe('when getting statistics', () => {
+
+        beforeEach( () => {
+            dockerMonitor = new DockerMonitorRewire(monitorOptions);
+        });
+
+        const cpuCores = os.cpus().length;
```
davidkel, line 175: If this test was run on a single core machine it would pass even if the code for cpu normalisation had the division removed. If possible we should stub out the use of os, the danger here is that it could break other tests that rely on os and expect os to actually work.in rewire there are some features such as revert which looks like they can change things back. I think you will want to call revert in the after block of the whole test file

packages/caliper-core/lib/manager/monitors/monitor-docker.js
davidkel, line 365: looks like you didn't make the suggested change ?

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+    const SystemInformationStub = {
+        dockerContainers : sinon.stub(),
+        dockerContainerStats : sinon.stub()
+    };
+    SystemInformationStub.dockerContainers.returns([{
+        id:"118",
+        name:"peer0.org1.example.com"
+    }]);
+    SystemInformationStub.dockerContainerStats.withArgs("118").returns(dockerStats);
+    const revertSistemInfo =DockerMonitorRewire.__set__("SystemInformation", SystemInformationStub)
```
davidkel, line 94: Sistem-->System

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+
+    describe('when retrieving container statistics', () => {
+
+        //memory test values
+        const mem_usage = dockerStats.memory_stats.usage - dockerStats.memory_stats.stats.cache;
+        const mem_percent = mem_usage / dockerStats.mem_limit;
+
+        it('should set isReading to false', async () => {
+            let dockerMonitor = new DockerMonitorRewire(monitorOptions);
+            await dockerMonitor.findContainers().then(async () => {
```
davidkel, line 167: Any reason why you don't just do
```
await dockerMonitor.findContainers();
await dockerMonitor.readContainerStats();
dockerMonitor.isReading.should.be.false;
```
?

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+                await dockerMonitor.readContainerStats()
+            }).then(async () => {
+                dockerMonitor.isReading.should.be.false;
+            });
+        });
+
+        it('should read and add stats for each target container', async () => {
+            let dockerMonitor = new DockerMonitorRewire(monitorOptions);
+            await dockerMonitor.findContainers().then(async () => {
```
davidkel, line 176: see above comment. you shouldn't have to use `.then` at all

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+
+        const monitorOptionsNotCpuNorm = {
+            interval: 5,
+            containers: ["peer0.org1.example.com"],
+            cpuUsageNormalization: false
+        };
+
+        it('should normalize cpuUsage if cpuUsageNormalization property is set to true', async () => {
+            let dockerMonitor = new DockerMonitorRewire(monitorOptionsCpuNorm);
+            await dockerMonitor.findContainers().then(async () => {
```
davidkel, line 211: shouldn't need to use `.then`

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+            }).then(async () => {
+                return await dockerMonitor.getStatistics()
+            }).then(function(result) {
+                result.resourceStats[0].get("CPU%(max)").should.equal(cpu_percent / 2); // 2 -> number of cpus used in this test
+            });
+        });
+
+        it('should not normalize cpuUsage if cpuUsageNormalization property not provided', async () => {
+            let dockerMonitor = new DockerMonitorRewire(monitorOptions);
+            await dockerMonitor.findContainers().then(async () => {
```
davidkel, line 222: Shouldn't need to use `.then`

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+            }).then(async () => {
+                return await dockerMonitor.getStatistics()
+            }).then(function(result) {
+                result.resourceStats[0].get("CPU%(max)").should.equal(cpu_percent);
+            });
+        });
+
+        it('should not normalize cpuUsage if cpuUsageNormalization property is set to false', async () => {
+            let dockerMonitor = new DockerMonitorRewire(monitorOptionsNotCpuNorm);
+            await dockerMonitor.findContainers().then(async () => {
```
davidkel, line 233: shouldn't need to use `.then`

packages/caliper-core/lib/manager/monitors/monitor-docker.js
fraVlaca: I forgot to add the changes to the main docker mon itor file and only pushed the one of the test file my bad

packages/caliper-core/test/manager/monitors/monitor-docker.js
```diff
@@ -0,0 +1,243 @@
...
+    osStub.cpus.returns(returnValueCpus);
+    const revertOS = DockerMonitorRewire.__set__("os", osStub)
+
+
+    //create System information stub
+    const SystemInformationStub = {
+        dockerContainers : sinon.stub(),
+        dockerContainerStats : sinon.stub()
+    };
+    SystemInformationStub.dockerContainers.returns([{
```
davidkel, line 89: this might need to be `resolves` rather than `returns`

fraVlaca: yup, you are right thanks

davidkel (approved):
LGTM

Issue #1141, opened 2021-05-17
can't catch the error of commitTimeOut , because of the unit is not uniform


workhard-12:
When we submit a tx , the unit of "commitTimeOut" should be seconds ,  but the caliper use milliseconds.

workhard-12:
packages\caliper-fabric\lib\connector-versions\v2\FabricGateway.js
- async _createGatewayWithIdentity(mspId, aliasName, wallet) 
- constructor(connectorConfiguration, workerIndex, bcType)

davidkel:
This is mainly a documentation issue as internally the code actually considers this to be seconds not milliseconds, however the default value of 60000 is assuming to be 60 seconds so the default value should change

davidkel:
Docs have been updated and code base has changed the default from 60,000 to 60. In the mean time you can override the default with
```
--caliper-fabric-timeout-invokeorquery 60
```
when you launch caliper

PR #1230, opened 2022-02-21
Correct timeout defaults

davidkel:
This is the apparent cause of a caliper run hang where it continues to
run with unfinished txns. It's actually waiting 60,000 seconds for txns
to finish which won't finish as it looks like the node sdk misses the
events.

code part of #1141 also addresses in part


aklenik (approved):
LGTM 👍

PR #1231, opened 2022-02-21
Update node versions for caliper

### LINKED ISSUE #1125: Caliper does not support nodejs 8.x
Caliper does not support nodejs 8.x, so documents should be updated.

For example, the following code use node 10 api (fs.promises)
https://github.com/hyperledger/caliper/blob/master/packages/caliper-fabric/lib/identity-management/IdentityManager.js#L17

davidkel:
closes #1125 


aklenik (approved):
👍

PR #1235, opened 2022-02-22
Remove the legacy fabric connectors

davidkel:
This is mainly a deletion exercise with the following code changes
1. simplication of which connector to load
2. only create a connector instance if required for example
flow-only-start and flow-only-end don't need to have a connector created


packages/caliper-fabric/test/FabricConnectorFactory.js
```diff
@@ -207,15 +171,13 @@ describe('A Fabric Connector Factory', () => {
...
-        mockery.registerMock('fabric-client', {});
-        mockery.registerMock('fabric-client/package', {version: '3.0.0'});
         ConfigUtil.set(ConfigUtil.keys.NetworkConfig, path.resolve(__dirname, v2Config));
         ConfigUtil.set(ConfigUtil.keys.Fabric.Gateway.Enabled, true);
         await ConnectorFactory(1).should.be.rejectedWith(/Installed SDK version 3.0.0 did not match any compatible Fabric connectors/);
         mockery.deregisterAll();
     });
 
-    it('should throw an error version if not for legacy or the new connector', async () => {
+    it('should throw a generic error if network configuration version is1.0 or 2.0', async () => {
```
aklenik, line 180: The description is a bit unclear. Shouldn't it end with `version is not 2.0`

davidkel: Well spotted. I blame my keyboard, it's wireless and so may times it doesn't register keystrokes but it's ergonomic :-)

aklenik (approved):
LGTM 👍

PR #1288, opened 2022-03-24
Caliper terminates if prometheus is not available

### LINKED ISSUE #1267: Test runs should not fail if it's not possible to extract prometheus data
In the case of prometheus we can specify a monitor to extract data from prometheus, however if prometheus isn't available you get
```
Error: connect ECONNREFUSED 127.0.0.1:9090
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1159:16)
Emitted 'error' event on ClientRequest instance at:
    at Socket.socketErrorListener (_http_client.js:475:9)
    at Socket.emit (events.js:400:28)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 9090
}
```
And caliper terminates. 

This is really annoying because now I can't use a benchmark which defines a monitor for prometheus if I choose not to have prometheus running.

It should not fail the run, just report a warning or at least have an option that stops it from failing the run

davidkel:
This is due to the error event not being correctly captured when a
request is made to prometheus

Also added some extra code to output a warning and stop trying to do any
more queries for the round.

It won't stop it for all rounds but checks on every round.

closes #1267

### DIFF
--- a/packages/caliper-core/lib/common/prometheus/prometheus-query-client.js
+++ b/packages/caliper-core/lib/common/prometheus/prometheus-query-client.js
@@ -149,10 +149,11 @@ class PrometheusQueryClient {
                         resolve();
                     }
                 });
-                res.on('error', err => {
-                    Logger.error(err);
-                    reject(err);
-                });
+            });
+
+            req.on('error', err => {
+                Logger.error(err);
+                reject(err);
             });
 
             req.end();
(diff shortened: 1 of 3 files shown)


aklenik (approved):
LGTM

PR #1298, opened 2022-04-05
Peer Gateway: Added Connector Selector implementation modifications to support the new Peer Gateway Connector

fraVlaca:
- added logic to pick new Peer Gateway connector in the connector selector in FabriConnectorFactory.js 
- added 2.4 fabric SUT version with binding packages in .config.yaml
- added last fixes for the Peer Gateway connector implementation

### DIFF
--- a/packages/caliper-fabric/lib/FabricConnectorFactory.js
+++ b/packages/caliper-fabric/lib/FabricConnectorFactory.js
@@ -19,53 +19,87 @@ const ConnectorConfigurationFactory = require('./connector-configuration/Connect
 const Logger = CaliperUtils.getLogger('FabricConnectorFactory');
 const semver = require('semver');
 
-const NEW_V1_NODE_CONNECTOR = './connector-versions/v1/FabricNonGateway.js';
-const NEW_V1_GATEWAY_CONNECTOR = './connector-versions/v1/FabricGateway.js';
-const NEW_V1_WALLET_FACADE_FACTORY = './connector-versions/v1/WalletFacadeFactory.js';
+const V1_NODE_CONNECTOR = './connector-versions/v1/FabricNonGateway.js';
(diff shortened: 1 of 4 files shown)


packages/caliper-fabric/lib/FabricConnectorFactory.js
```diff
@@ -26,46 +26,73 @@ const NEW_V1_WALLET_FACADE_FACTORY = './connector-versions/v1/WalletFacadeFactor
...
+                    throw new Error('Caliper currently only supports Fabric gateway based operation using Fabric-SDK 1.4.2 and higher. Please retry with a different SDK binding');
+                }
             }
+        } else if (semver.satisfies(version, '=2.x')) {
+            Logger.info(`Initializing gateway connector compatible with installed SDK: ${version}`);
+            connectorPath = NEW_V2_GATEWAY_CONNECTOR;
+            walletFacadeFactoryPath = NEW_V2_WALLET_FACADE_FACTORY;
+        } else {
+            throw new Error(`Installed SDK version ${version} did not match any compatible Fabric connectors`);
+        }
```
davidkel, line 87: We should just do a return here and then we can remove the following `else if` line, just a comment to say it must be a peer-gateway sdk

packages/caliper-fabric/lib/connector-versions/peer-gateway/WalletFacade.js
```diff
@@ -50,6 +50,7 @@ class WalletFacade extends IWalletFacade {
                 privateKey
             },
             mspId,
+            type: 'X.509',
```
davidkel, line 53: You don't actually need this line, so can be removed

davidkel (changes requested):
Some minor comments, but looking good.

davidkel (changes requested):
mainly just typos to correct

davidkel (approved):
LGTM

PR #1306, opened 2022-04-08
Fixes for default.yaml in packages/caliper-core/lib/common/config/ and updates for the fabric Channel operations for the v1 fabric connector

fraVlaca:
edited defaul.yaml (updated comments and deleted sllep after option for fabric) + updated the fabric Chaincode operations for the v1 fabric connector


davidkel (changes requested):
Wrong file updated

davidkel (approved):
LGTM

PR #1314, opened 2022-04-14
Change monitor intervals from milliseconds to seconds

davidkel:
For the fabric tests the monitor intervals were set to single
millisecond values such as 3, 4, 5. This adds a massive overhead to the
running of the integration tests and isn't needed, so changing them to
be seconds instead.


aklenik (approved):
This is probably a leftover bug from when `interval` was considered as seconds.
Maybe we should look at unifying the duration specifications across Caliper with something like this: https://www.npmjs.com/package/parse-duration

PR #1329, opened 2022-05-06
add peers property support to fabric network config

davidkel:
Add ability to list 1 or more fabric peers in an organisation that can
be used to discover the fabric network. The new Peer Gateway connector
will only ever use the first one (Issue exists to allow for more gateway
peers to be used as load balancers)

This means connection profiles are not needed anymore reducing the
network configuration to a single file (ideally you would embed the
tlsCACert information as well

This will work for
- bound fabric 1.4 using gateway-enabled
- bound fabric 2.2
- bound fabric 2.4

It will not work with bound fabric 1.4 without gateway-enabled and the
connector will report a failure saying that it cannot work with a
dynamic connection profile (there is an issue to enable dynamic profile
support for this but the 1.4 connector is legacy and not worth investing
in)

The implementation works by generating a dynamic connection profile from
the peer information as all the connectors work by working to an
interface which abstracts the required information but the
implementation gets that information by processing an in memory version
of the connection profile and doesn't make sense to have 2 different in
memory representations.

Documentation still to be provided


aklenik (approved):
LGTM 👍

PR #1332, opened 2022-05-12
Upgraded node-sdk binding for fabric-v1-lts from 1.4.19 to 1.4.20

fraVlaca:
Node SDK v1.4.20 relaxes the Node version check to >=10.13.0, so should be happy with any later Node version, even though only the LTS releases are “supported”

This upgrade will help remove all the warnings about node versions on caliper.

Signed-off-by: fraVlaca <ocsenarf@outlook.com>


davidkel (approved):
LGTM

PR #1346, opened 2022-05-17
Replaced bit.ly link with reference to bootstrap.sh

fraVlaca:
Replaced bit.ly with reference bootstrap.sh as per the build (but without the -k option)
Signed-off-by: fraVlaca <ocsenarf@outlook.com>


davidkel (approved):
LGTM

PR #1363, opened 2022-05-31
Fix #536 Update error messages for 'caliper-ethereum' package

### LINKED ISSUE #536: Make logging consistent
The current logging situation is a little bit ad-hoc. There is no enforced standard, and we would benefit greatly from a serviceability and usability point if we were to enforce logging standards across the Caliper code base.

We should be logging at debug level:
- [ ] entry/exit from every method, including passed values

We should be logging at error level
- [ ] every error thrown through the code

We should be logging at info level:
- [ ] caliper lifecycle items
- [ ] round information and basic statistics (we do this already)

Tezas-6174:
- Partially fixes # 536. 
- The error messages are clear and consistent.
- All the error messages i.e. at debug level, error level, and info level are checked and updated.

### DIFF
--- a/packages/caliper-ethereum/lib/ethereum-connector.js
+++ b/packages/caliper-ethereum/lib/ethereum-connector.js
@@ -119,10 +119,10 @@ class EthereumConnector extends ConnectorBase {
                 try {
                     if (privacy) {
                         contractInstance = await self.deployPrivateContract(contractData, privacy);
-                        logger.info('Deployed private contract ' + contractData.name + ' at ' + contractInstance.options.address);
+                        logger.info(`Deployed contract ${contractData.name} at ${contractInstance.options.address}`);
                     } else {
                         contractInstance = await self.deployContract(contractData);
-                        logger.info('Deployed contract ' + contractData.name + ' at ' + contractInstance.options.address);
+                        logger.info(`Deployed contract ${contractData.name} at ${contractInstance.options.address}`);
                     }
                 } catch (err) {
                     reject(err);
@@ -244,7 +244,7 @@ class EthereumConnector extends ConnectorBase {
 
         const onFailure = (err) => {
             status.SetStatusFail();
-            logger.error('Failed tx on ' + request.contract + ' calling method ' + request.verb + ' nonce ' + params.nonce);
+            logger.error(`Failed tx on ${request.contract}; calling method: ${request.verb}; nonce: ${params.nonce}`);
             logger.error(err);
         };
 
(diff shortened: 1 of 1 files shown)


packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -302,7 +302,7 @@ class EthereumConnector extends ConnectorBase {
 
         const onFailure = (err) => {
             status.SetStatusFail();
-            logger.error('Failed private tx on ' + request.contract + ' calling method ' + request.verb + ' private nonce ' + 0);
+            logger.error(`Failed private tx on ${request.contract}; calling method: ${request.verb}; private nonce: ${0}`);
```
davidkel, line 305: not sure why the original log entry had `+ 0` but it can't be replaced with `${0}` here. I think it might be the value of `sender.nonce` if set and literal `0` otherwise

packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -402,8 +402,8 @@ class EthereumConnector extends ConnectorBase {
             if (txRcpt.status === '0x1') {
                 return new web3.eth.Contract(contractData.abi, txRcpt.contractAddress);
             } else {
-                logger.error('Failed private transaction hash ' + txHash);
-                throw new Error('Failed private transaction hash ' + txHash);
+                logger.error(`Failed private transaction hash ${txHash}`);
```
davidkel, line 405: As this message is duplicated, better to set it as a variable and pass it into `logger.error` and `new Error`

davidkel (changes requested):
a few minor comments

packages/caliper-ethereum/lib/ethereum-connector.js
Tezas-6174: Is it ``privacy.nonce`` or sender.nonce``?
Shall I add a condition to check if ``sender.nonce`` is set ? If not then I'll set it to 0 and then will replace the ``0`` above with the value of ``sender.nonce``.
Is it fine to do it this way?

davidkel: My recommendation would be to not change the message around the nonce. If you change it to sender.nonce then you will need to write unit tests and prove that it works. So suggest you just do
```
logger.error(`Failed private tx on ${request.contract}; calling method: ${request.verb}; private nonce: ` + 0);
```

Tezas-6174: Ok, I'll change it.

packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -402,8 +402,9 @@ class EthereumConnector extends ConnectorBase {
             if (txRcpt.status === '0x1') {
                 return new web3.eth.Contract(contractData.abi, txRcpt.contractAddress);
             } else {
-                logger.error('Failed private transaction hash ' + txHash);
-                throw new Error('Failed private transaction hash ' + txHash);
+                let msg = `Failed private transaction hash ${txHash}`;
```
davidkel, line 405: as msg never changes it should be `const` not `let`

Tezas-6174: Oh, yes, sorry it happened by mistake. Updated it.

davidkel (approved):
LGTM

PR #1364, opened 2022-06-08
Port the unit test workflow to GitHub Actions

### LINKED ISSUE #1359: Create Github Actions pipeline to perform the equivalent of the UnitTest stage in Azure pipelines
This pipeline should replicate the following 
- [ ] runs under both node 14 and node 16
- [ ] Check correct usage of Caliper package names
- [ ] Install project dependencies
- [ ] Check the version consistency of subpackages
- [ ] Run unit tests

This will run in parallel with the current Azure Pipelines

CaptainIRS:
In this PR:
* The unit test workflow which exists in Azure Pipelines is ported over to GitHub Actions
* A badge is added to the README for the workflow

The migration is complete. Optimizations like cache management will be tackled later.

Closes #1359


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        path: './node_modules'
+        key: ${{ runner.os }}-node-${{ matrix.node-version }}-cache-${{ hashFiles('package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-${{ matrix.node-version }}-cache-
+    - name: Install project dependencies
+      if: steps.cache-npm.outputs.cache-hit != 'true'
+      run: npm install
+    - name: Bootstrap lerna
+      run: npm run bootstrap
```
aklenik, line 35: This step also calls `npm install` for all packages managed in the repository. And they're more heavyweight installs than the root `npm install` (which installs 3 dev packages). And after hoisting (performed by lerna), the root `node_modules` will contain the common dependencies of every Caliper package, and the `node_modules` of individual packages will contain their unique dependencies. 

So could we somehow cache every `node_modules` path?

CaptainIRS: I think this shouldn't be a problem as the `node_modules` folder is cached only after the whole workflow is completed and lerna's bootstrap step installs the dependencies only in the root `node_modules` folder. (I can confirm this as the size of the root `node_modules` folder grew from 106MB to 774MB after the bootstrap step when testing locally. Also I have observed that the time taken for the bootstrap step is cut down from 4m 36s in the first run to 20s in a cached run of GitHub Actions in a test PR I have created in my fork)

davidkel: Lerna hoists packages up to the root node_modules only if it can so you will see some node modules still residing under the different packages in caliper. The package-lock.json file only contains the description of the packages required for bootstrapping the caliper repo and not for the node modules required for each package, so do you know how much dependency there is on needing the package-lock.json file ? It may be worth not including that file for now and looking into how we utilise package-lock.json files to their full potential within a lerna environment. There is a separate issue to cover this.

package-lock.json
```diff
@@ -0,0 +1,18428 @@
+{
```
davidkel, line 1: We should remove this file for now

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+        node-version: [14.x, 16.x]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Check correct usage of Caliper package names
+      run: ./scripts/check-package-names.sh
+    - name: Attempt to use cached dependencies
```
davidkel, line 23: we should remove caching for now

davidkel (changes requested):

.github/workflows/unit-tests.yml
CaptainIRS: I've just noticed the `node_modules` folder in the sub-packages. Thanks for the info!

Now thinking of caching with lerna, it seems to be a lot trickier. I will try to write a doc with some proposed ways of solving this issues and some analysis of pros and cons of each solution.

davidkel (approved):
LGTM

PR #1376, opened 2022-06-20
Port the integration test workflow to GitHub Actions

### LINKED ISSUE #1366: Port the integration tests to github actions
Port the integration tests to github actions as is , ie without trying to improve it just to get it migrated

CaptainIRS:
In this PR:

* The integration test workflow which exists in Azure Pipelines is ported over to GitHub Actions.
* A badge is added to the README for the workflow.
* The migration is complete. Optimizations like cache management will be tackled later.
* The `package-lock.json` file which was earlier removed from `.gitignore` is added back.

Closes #1366


davidkel:
@CaptainIRS This is looking good, but it looks like the unit tests and the integration tests run in parallel. I think it would be better if the integrations tests are only run in the unit test stage passes. If the unit tests fail we probably don't want the longer integration tests run until the unit tests are fixed.

davidkel (changes requested):
I think we want to replicate what the azure pipelines currently do which is run the unit tests first and if they pass then move on to running the integration tests

davidkel (approved):
LGTM

PR #1387, opened 2022-06-22
Cache node modules across CI workflows

### LINKED ISSUE #1374: Investigate the ability to cache the npm node modules required for the unit tests/Integration tests to run
Caliper installs quite a few npm modules using lerna which results in several node_modules directories

We should investigate what would be required to cache these for subsequent builds but also be able to invalidate part or all of the cache if a node module dependency is updated.

CaptainIRS:
In this PR:
* The `node_modules` folder is cached across workflows using a method inspired by the workflow in [jupyterlab/extension-examples](https://github.com/jupyterlab/extension-examples/blob/master/.github/workflows/main.yml)

Closes #1374

### DIFF
--- a/.build/benchmark-integration-test-direct.sh
+++ b/.build/benchmark-integration-test-direct.sh
@@ -17,9 +17,6 @@
 set -e
 set -o pipefail
 
-# Bootstrap the project again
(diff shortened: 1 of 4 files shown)


CaptainIRS:
Superseded by #1406

PR #1393, opened 2022-06-28
Add DCI linting to CI of PRs

aklenik:
As recommended by the project update template and https://github.com/petermetz/gh-action-dci-lint
Let's see whether the remnant phrases in the changelog and third-party artifacts will trigger the alarm before making this a branch protection rule.

### DIFF
--- a/packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js
+++ b/packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js
@@ -217,7 +217,7 @@ describe('When using a PrometheusPushTxObserver', () => {
         await prometheusPushTxObserver.deactivate();
 
         // Values should be zero, or empty 
-        // Ref: https://github.com/siimon/prom-client/  file test/counterTest.js
+        // Ref: https://github.com/siimon/prom-client/blob/721829cc593bb7da28ae009985caeeacb4b59e05/test/counterTest.js
         const txSubmitted = await prometheusPushTxObserver.counterTxSubmitted.get();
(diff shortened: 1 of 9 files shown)


packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js
```diff
@@ -216,7 +216,8 @@ describe('When using a PrometheusPushTxObserver', () => {
 
         await prometheusPushTxObserver.deactivate();
 
-        // Values should be zero, or empty (https://github.com/siimon/prom-client/blob/master/test/counterTest.js)
+        // Values should be zero, or empty 
+        // Ref: https://github.com/siimon/prom-client/  file test/counterTest.js
```
davidkel, line 220: this url doesn't look valid. If we can't use the real url then we should delete the line

aklenik: It's split into repo and file parts. I could structure it better if needed. But the entire file URL contains `master`...
Or we can remove it entirely

davidkel (changes requested):
just a couple of minor comments

packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js
aklenik: @davidkel, could be structured like this:
`See the "test/counterTest.js" file in https://github.com/siimon/prom-client/`

davidkel (approved):
LGTM

PR #1400, opened 2022-07-04
Add skeleton for development and proposal docs

aklenik:
Draft PR showcasing the envisioned structure of the dev docs.

### DIFF
--- a/dev-docs/README.md
+++ b/dev-docs/README.md
@@ -1,0 +1,33 @@
+# Hyperledger Caliper Developer Documentation
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


davidkel (commented):
I think we need to say somewhere that anything merged into proposals means it has been accepted. Once implemented they should be moved into an appropriate place in dev-docs
Do we need 2 generate-figure.sh scripts ?
I love the idea of using plantuml, something I have used before as well
If a proposal has multiple figures maybe we need some easy way of distinguishing them for a proposal so that it makes it easy to move ?
But I really like the idea here

aklenik:
@davidkel, thanks for the insights! I've updated the readme and kept only one fig-gen script.

davidkel (approved):
LGTM

PR #1404, opened 2022-07-12
Vscode extension

eravatee:
The primary body of work to setup a vscode extension for caliper. 
It introduces a json schema to assist users in creating a runtime configuration for caliper


packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "caliper-vscode-extension",
```
aklenik, line 2: Since extensions are stored directly on the vscode marketplace, we should use a different naming convention here (not like the other packages for NPM). I'd suggest: `hyperledger.caliper-extension`

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+	"displayName": "caliper-vscode-extension",
```
aklenik, line 3: This should be a human-friendly name, like `Hyperledger Caliper Integration`

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+	"description": "",
```
aklenik, line 4: This should be filled with a single, short, descriptive sentence about the extension. Look up some extensions in vscode to get a sense of what kinds of descriptions should be used.

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+	"version": "0.0.1",
```
aklenik, line 5: Every Caliper package version is in sync with each other to give a single "global" version for Caliper. So this "package" should also match the current Caliper version, which is `0.5.1-unstable`

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+	"engines": {
+		"vscode": "^1.68.0"
+	},
+	"categories": [
+		"Other"
+	],
+	"activationEvents": [
+        "onCommand:caliper-vscode-extension.helloWorld"
```
aklenik, line 13: Sample events can be removed.

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+	],
+	"main": "./out/extension.js",
+	"contributes": {
+		"commands": [
+			{
+				"command": "caliper-vscode-extension.helloWorld",
```
aklenik, line 19: Sample commands can be removed.

packages/caliper-vscode/extension/package.json
```diff
@@ -0,0 +1,45 @@
...
+		"vscode": "^1.68.0"
+	},
+	"categories": [
+		"Other"
+	],
+	"activationEvents": [
+        "onCommand:caliper-vscode-extension.helloWorld"
+	],
+	"main": "./out/extension.js",
+	"contributes": {
```
aklenik, line 16: Schema definitions should be added here declaratively, and not from code.
See: https://code.visualstudio.com/api/references/contribution-points#contributes.jsonValidation
Also, schema URLs should point to the local files in the package, so the extension is self-contained (and evolves together with the schemas).

I'd propose the following naming conventions for now (same for the .yml and .json extensions), and we can discuss them later on the contributor channel in detail:
* Benchmark configuration file: `*.calbench.yaml`
* Runtime configuration file: `*.calrun.yaml`
* Fabric network configuration file: `*.fabric.calnet.yaml`
* Ethereum network configuration file: `*.eth.calnet.yaml`

packages/caliper-vscode/extension/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
```
aklenik, line 1: This should extend the available Node14 base:
* How-to: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases
* Needed package: @tsconfig/node14
* Content of base file: https://github.com/tsconfig/bases/blob/main/bases/node14.json

packages/caliper-vscode/extension/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Change Log
```
aklenik, line 1: The changelog is currently maintained at the repo level, so this file is not needed.

packages/caliper-vscode/extension/.eslintrc.json
```diff
@@ -0,0 +1,24 @@
+{
```
aklenik, line 1: You should reuse the eslintrc file content from one of the other packages, like caliper-core, so it's in sync with the other packages

packages/caliper-vscode/artifacts/benchmark.json
```diff
@@ -0,0 +1,313 @@
...
+                                "properties": {
+                                    "label": {
+                                        "description": "A short name of the rounds, usually corresponding to the types of submitted TXs.",
+                                        "type": "string"
+                                    },
+                                    "txNumber": {
+                                        "description": "The number of TXs Caliper should submit during the round.",
+                                        "type": "integer"
+                                    },
+                                    "rateControl": {
```
aklenik, line 57: Depending on the value of `type`, the `opts` object can be different. I'd suggest exploring conditional schemas to handle such cases flexibly: https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else
This will require a newer JSON schema version, but vscode supports even the newest draft, so I'd suggest that every schema should use that newest draft.

aklenik (changes requested):
Besides the comments embedded in the files, I have the following general remarks:
1. The package directory name should be `caliper-vscode-extension`
2. This directory should be the package's root, so no other `extension` subdirectory is needed.
3. The schema files should go into the `caliper-vscode-extension/artifacts/schemas` directory. We might have different types of artifacts later on.
4. The schema file names should be more descriptive (their length doesn't matter, users will rarely type it themselves):
    1. `benchmark.json` -> `hyperledger-caliper-benchmark-confguration-schema.json`
    2. `default.json` -> `hyperledger-caliper-runtime-configuration-schema.json`
    3. `fabric.json` -> `hyperledger-caliper-fabric-network-configuration-schema.json`
    4. `ethereum-besu.json` -> `hyperledger-caliper-ethereum-network-configuration-schema.json`
5.  You should check whether the current configuration files used by the CI pipeline (in `packages/caliper-tests-integration`) pass the schema validation. Those configs cover a lot of different use cases, so until we don't have proper unit tests for the schemas, such smoke tests can help.
6. Since the extension only registers schemas, we won't have any TS source files in the PR for now. So every unrelated sources can be removed from the PR.
7. For starters, I suggest keeping only the runtime configuration schema for this PR. The other schemas are more complicated, and even the runtime schema is incomplete regarding descriptions.

packages/caliper-vscode/extension/package.json
eravatee: This type of naming syntax is not acceptable with vscode. I went with hyperledger-caliper-extension

packages/caliper-vscode-extension/artifacts/json-schema/hyperledger-caliper-fabric-network-configuration.schema.json
```diff
@@ -0,0 +1,350 @@
+{
```
aklenik, line 1: For now, let's remove this schema from the PR. Let's focus only on the runtime configuration schema.

packages/caliper-vscode-extension/artifacts/json-schema/hyperledger-caliper-ethereum-network-configuration.schema.json.json
```diff
@@ -0,0 +1,129 @@
+{
```
aklenik, line 1: For now, let's remove this schema from the PR. Let's focus only on the runtime configuration schema.

packages/caliper-vscode-extension/artifacts/json-schema/hyperledger-caliper-benchmark-configuration.schema.json
```diff
@@ -0,0 +1,313 @@
+{
```
aklenik, line 1: For now, let's remove this schema from the PR. Let's focus only on the runtime configuration schema.

packages/caliper-vscode-extension/artifacts/json-schema/hyperledger-caliper-runtime-configuration.schema.json
```diff
@@ -0,0 +1,543 @@
+{
+    "$schema": "https://json-schema.org/draft-06/schema",
+    "$ref": "#/definitions/ConfigurationHelper",
+    "definitions": {   
+        "ConfigurationHelper": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "caliper": {
```
aklenik, line 9: Every root/middle/leaf property must have a user-friendly description, summarizing the property(group) as precisely as possible

aklenik (changes requested):
There are still some unresolved remarks or solutions that don't follow the suggested approaches (eslint, tsconfig contents, left-over out-of-scope schemas). Please resolve these systematically.

aklenik (approved):

PR #1406, opened 2022-07-17
Cache node modules across CI workflows

### LINKED ISSUE #1374: Investigate the ability to cache the npm node modules required for the unit tests/Integration tests to run

CaptainIRS:
In this PR:
* The integration tests workflow is refactored to be DRY
* The `node_modules` folder is cached across workflows
* The SUT binding is cached across integration test runs

Closes #1374

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -7,11 +7,13 @@ env:
   NPM_VERSION: 8.15.0
 
 jobs:
-  fabric-integration-tests:
-    name: Fabric Integration Tests
+  integration-tests:
+    name: Integration Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
+        adaptor: [fabric, ethereum, besu, fisco-bcos, generator]
         node-version: [14.x, 16.x]
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


CaptainIRS (commented):
Hi @davidkel, this PR is now ready for review.

packages/caliper-tests-integration/fabric_tests/run.sh
```diff
@@ -29,10 +29,13 @@ cd ${DIR}
 
 # bind during CI tests, using the package dir as CWD
 # Note: do not use env variables for binding settings, as subsequent launch calls will pick them up and bind again
+FABRIC_VERSION=1.4.20
```
davidkel, line 32: can we leave this a 1.4 ? then it will always test the latest 1.4 version without any changes to the integration tests

packages/caliper-tests-integration/fabric_tests/run.sh
```diff
@@ -81,18 +84,14 @@ if [[ ${rc} != 0 ]]; then
...
-# UNBIND SDK, using the package dir as CWD
-# Note: do not use env variables for unbinding settings, as subsequent launch calls will pick them up and bind again
-if [[ "${BIND_IN_PACKAGE_DIR}" = "true" ]]; then
-    pushd $SUT_DIR
-    ${CALL_METHOD} unbind --caliper-bind-sut fabric:1.4
-    popd
-fi
 # BIND with 2.2 SDK, using the package dir as CWD
 # Note: do not use env variables for unbinding settings, as subsequent launch calls will pick them up and bind again
+FABRIC_VERSION=2.2.14
```
davidkel, line 89: Similar here, could we leave it as 2.2 ?

packages/caliper-tests-integration/fisco-bcos_tests/run.sh
```diff
@@ -24,7 +24,7 @@ cd "${DIR}"
 # Note: do not use env variables for binding settings, as subsequent launch calls will pick them up and bind again
 if [[ "${BIND_IN_PACKAGE_DIR}" = "true" ]]; then
     pushd $SUT_DIR
-    ${CALL_METHOD} bind --caliper-bind-sut fisco-bcos:latest
+    ${CALL_METHOD} bind --caliper-bind-sut fisco-bcos:2.0.0
```
davidkel, line 27: do we need to change this ? just leave it as latest ?

packages/caliper-tests-integration/besu_tests/run.sh
```diff
@@ -26,7 +26,7 @@ npm i
 # Note: do not use env variables for binding settings, as subsequent launch calls will pick them up and bind again
 if [[ "${BIND_IN_PACKAGE_DIR}" = "true" ]]; then
     pushd $SUT_DIR
-    ${CALL_METHOD} bind --caliper-bind-sut besu:latest
+    ${CALL_METHOD} bind --caliper-bind-sut besu:1.4
```
davidkel, line 29: do we have to change this ? just leave it as latest ? If it's a clash problem, then we could include the sut name in the directory structure. Only fabric has different connector versions to be tested

packages/caliper-tests-integration/ethereum_tests/run.sh
```diff
@@ -24,7 +24,7 @@ cd "${DIR}"
 # Note: do not use env variables for binding settings, as subsequent launch calls will pick them up and bind again
 if [[ "${BIND_IN_PACKAGE_DIR}" = "true" ]]; then
     pushd $SUT_DIR
-    ${CALL_METHOD} bind --caliper-bind-sut ethereum:latest
+    ${CALL_METHOD} bind --caliper-bind-sut ethereum:1.3
```
davidkel, line 27: do we have to change this could it be left as latest ?

.github/workflows/integration-tests.yml
```diff
@@ -21,83 +23,37 @@ jobs:
...
-    - name: Generator Integration Test
-      run: .build/benchmark-integration-test-direct.sh
+        path: /tmp/sut
+        key: ${{ format('{0}-sut-cache-{1}-{2}', matrix.node-version, matrix.adaptor, hashFiles(format('packages/caliper-tests-integration/{0}_tests/run.sh', matrix.adaptor))) }}
+        restore-keys: |
+          ${{ matrix.node-version }}-sut-cache-${{ matrix.adaptor }}-
+    - name: Install project dependencies
+      if: steps.cache-npm.outputs.cache-hit != 'true'
+      run: npm ci
+    - name: Create SUT binding
```
davidkel, line 47: This should say Create directory to store SUT bindings

packages/caliper-tests-integration/run-tests.sh
```diff
@@ -25,6 +25,8 @@ if [[ "${BENCHMARK}" = "" ]]; then
     exit 1
 fi
 
+mkdir -p $SUT_DIR
```
davidkel, line 28: do we need this if it's part of the ci pipeline which Creates the SUT binding ?

davidkel (changes requested):
Only minor comments, we will still need the .build\benchmark-integration-test-direct.sh script so that devlopers can run integration builds locally. Before I just used to look at the ci pipelines to see how to do a local build but with the caching I think we need to document this script in the contributor guidelines  now as it provides an easy way to run the integration tests.
Does that sound ok or is there an alternative ?

packages/caliper-tests-integration/fabric_tests/run.sh
CaptainIRS: The problem with leaving it at latest is how do we invalidate the cache. Even when a new version is released (like 1.4.21), our CI pipeline will still be using the older cached version, giving a false impression that the tests are passing against the latest Fabric 1.4. If we were explicit about the versions used, I think it would be easy to spot if later versions cause issues with the codebase.

CaptainIRS (commented):
> Only minor comments, we will still need the .build\benchmark-integration-test-direct.sh script so that devlopers can run integration builds locally. Before I just used to look at the ci pipelines to see how to do a local build but with the caching I think we need to document this script in the contributor guidelines now as it provides an easy way to run the integration tests. Does that sound ok or is there an alternative ?

The script would still remain in the codebase for developers to test their contributions. I have opted to not use it in the CI pipeline as it also includes an `npm install` step and also deletes the SUT directory, which may be useful for a developer but is not suitable for the CI pipeline. Also the usage of the script has already been documented in https://hyperledger.github.io/caliper/v0.5.0/contributing/#testing-the-code.

packages/caliper-tests-integration/fabric_tests/run.sh
davidkel: So the cache is persistent across actions then ? if so then I see what you mean. However we then have a problem with 2.4 as we don't have a version number for it like 1.4.20, 2.2.14. 
It would be a bit of a pain to update the version number in caliper and remember to also upgrade the integrations tests and that was what we wanted to avoid.
1.4.20 and 2.2.14 aren't likely to be increased now, so maybe could we not cache the 2.4 one. the bind of 2.4 would be quite quick compared to 1.4 and 2.2 as 2.4 is very light weight

CaptainIRS: Yes. I have taken the versions from https://github.com/hyperledger/caliper/blob/main/packages/caliper-cli/lib/lib/config.yaml, however there isn't a minor version for 2.4 as of now, so I had to leave it as 2.4. Should we add a minor version in the config file or could I just remove the caching for 2.4 (I think removing caching for just 2.4 is going to be hacky)?

CaptainIRS: I was thinking of extracting the version numbers out of the script and passing it through the workflow, however I was thinking about two issues that it might raise:
1. The Fabric test script itself uses 3 versions of Fabric, so it would be a challenge to pass that to the script.
2. Even if we extract the version numbers to environment variables in the workflow, we still need to hard-code the version numbers in the script that developers use to run tests. So, that would only increase the number of places we need to change to update the version numbers.

davidkel: Also lets add some comments in explaining this ie what's cached and not cached and the use of the detailed version numbers in this file plus we should add comments to `packages\caliper-cli\lib\lib\config.yaml` as well otherwise I can see that this could get missed if the config.yaml file was updated but entries in the integration tests weren't

packages/caliper-cli/lib/lib/config.yaml
```diff
@@ -28,6 +28,8 @@ settingProfiles:
 # Specifies the supported target platforms
 sut:
     # Specifies the available SDK bindings and their Node.js-specific install settings
+    # Note: The minor version for the SUT in integration tests need to be upgraded
```
davidkel, line 31: I think we need to provide more comprehensive information here

```suggestion
sut:
    # IMPORTANT: 
    # The Integration tests cache all the SUT bindings EXCEPT for the fabric:2.4 binding. Therefore
    # for those bindings if you want to update a binding with new version dependencies you need to 
    # create a new version and update the integration tests explicitly to reference this new version.
    # For example if the current version of the fabric 2.2 binding is 2.2.14 and you want to update
    # the binding to include a new level of dependencies, you must create a 2.2.15 binding and update
    # the integration tests to reference this. 
    # The exception is fabric:2.4 which isn't cached in the integration tests. This means you can modify
    # the 2.4 key without creating a new version and the integration tests will automatically pick it up.
```

packages/caliper-tests-integration/fabric_tests/run.sh
```diff
@@ -105,19 +106,15 @@ if [[ ${rc} != 0 ]]; then
...
-# Note: do not use env variables for unbinding settings, as subsequent launch calls will pick them up and bind again
-if [[ "${BIND_IN_PACKAGE_DIR}" = "true" ]]; then
-    pushd $SUT_DIR
-    ${CALL_METHOD} unbind --caliper-bind-sut fabric:2.2
-    popd
-fi
-
 # BIND with 2.4 SDK, using the package dir as CWD
 # Note: do not use env variables for unbinding settings, as subsequent launch calls will pick them up and bind again
+# Note: Fabric 2.4 binding is NOT cached in CI as it doesn't have the minor version in the cache key
```
davidkel, line 111:
```suggestion
# Note: Fabric 2.4 binding is NOT cached in CI. This binding is lightweight so doesn't take much time and allows the 2.4 binding to be modified in the config.yaml binding file
```

davidkel (changes requested):
I think we need to provide just some extra detail in the comments to explain what is going on

davidkel (approved):
LGTM

PR #1407, opened 2022-07-17
Add workflow to test npm packages

### LINKED ISSUE #1403: Publish: Automated tests for the npm packages before they are published to official npm repository
I think this should be done on the merge pipeline rather than the PR pipeline, but it should also be possible to run the scenario locally

There is an integration test in the integration package that used verdaccio to act as a local npm repository where npm packages could be published and then pulled to test that the packages were valid

We should bring this upto date and re-enable it, it will also be used to verify that moving to npm workspaces and including package-lock.json files will continue to work on standard installs of node 14 and node 16

This is part of the optimize theme as it ties in with caching npm modules and the pre-req of using npm workspaces as the replacement for lerna

CaptainIRS:
In this PR
* A new GitHub Actions workflow is added to test publishing of a package and the functioning of the published package using Verdaccio, with the help of scripts written previously

Closes #1403

### DIFF
--- a/.github/workflows/package-tests.yml
+++ b/.github/workflows/package-tests.yml
@@ -1,0 +1,37 @@
+name: Test Published Package
+
+on:
+  workflow_call:
+
+jobs:
+  package-tests:
+    name: Test Published Package
+    runs-on: ubuntu-latest
+    services:
+      verdaccio:
+        image: verdaccio/verdaccio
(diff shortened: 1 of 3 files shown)


davidkel:
Unsure why this was not reviewed and also why it was closed. Going to re-open

PR #1411, opened 2022-07-20
Remove channel and chaincode operations from Fabric v1 connector

### LINKED ISSUE #1204: Remove the ability for caliper to be able to create channels, install and instantiate chaincode for fabric
### Please share the technical limitation of Caliper that you encountered.

Caliper has the ability to create channels, install and instantiate chaincode however it is very limited
- Can only be used when bound to the 1.4 fabric SUT
- Only works with the legacy fabric lifecycle (which is now deprecated)
- Is not a complete implementation so not all scenarios are possible
- Isn't considered to be a problem that caliper should solve

### Please detail your feature idea that could alleviate the limitation.

The minimum we would need to do would be to update the documentation to not expose the options. Following on from that we would need to
- change the way integration tests are done for fabric as it relies on this to define the test network
- consider actually removing the code from the 1.4 connectors (No plan to remove the 1.4 connector yet)
- caliper benchmarks would need to rethink it's networks setup. This will be separate issue as to what we recommend for standing up and configuring a fabric network

### Please share some details about your use case if possible, and how the new feature would make Caliper a better performance benchmarking framework.

_No response_

### Please share any suggestions about the new feature's code/configuration API (using formatted YAML segments or pseudo-code).

_No response_

CaptainIRS:
In this PR:
* The ability for caliper to be able to create channels, install and instantiate chaincode for fabric is removed.

Closes #1204


davidkel (approved):
LGTM, great to finally remove this from caliper

PR #1414, opened 2022-07-25
Migrate generator tests to test network

### LINKED ISSUE #1254: generator integration tests need to be moved to fabric 2.2 or later
The caliper generator tests currently use fabric 1.4.8. We should replace this with either an explicit 2.2/2.4 system or use test-network from fabric-samples

### LINKED ISSUE #1413: Generator integration tests are failing
The generator integration tests are failing after the ability of Caliper to create channels and chaincode was removed. This is because generator integration tests still use Caliper's obsolete functionality. So, generator integration tests need to be migrated to test network.

CaptainIRS:
In this PR:
* Generator integration tests are migrated to test network so that they don't rely on obsolete Caliper features.
* Generator integration tests are moved to Fabric 2.2

Fixes #1254
Fixes #1413


davidkel (approved):
LGTM

PR #1421, opened 2022-07-30
Execute integration tests based on changes

### LINKED ISSUE #1375: Only run relevant tests on the integration builds depending on which package was changed
caliper runs several integration tests in parallel for
- ethereum
- fabric
- fisco-bcos
- besu
- generator

However we don't want to run them all if the code change only affects a specific component (we would then rely on a nightly build to run them all to pick up possible issues outside of a change being made)

We should detect changes at the package level to start with to determine what to test
- caliper-cli: fabric
- caliper-core: fabric (unless it's the config.yml file that controls bindings then it should be all)
- caliper-ethereum: ethereum, besu
- caliper-fabric: fabric (this could be split further based on change to a specific connector)
- caliper-fisco-bcos: fisco-bcos
- caliper-gui-dashboard/caliper-gui-server: don't run any integration tests
- caliper-publish: all
- caliper-test-integration: all
- generator-caliper: generator

CaptainIRS:
In this PR:
* The integration test workflow is modified to execute only tests that are relevant to the changes made

Closes #1375

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -7,13 +7,56 @@ env:
   NPM_VERSION: 8.15.0
 
 jobs:
+  changes:
+    name: Check for changes
+    runs-on: ubuntu-latest
+    outputs:
+      adaptor: ${{ steps.filter.outputs.changes }}
+    steps:
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          fabric:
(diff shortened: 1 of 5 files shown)


davidkel (changes requested):
It's looking good, but it seems there is an edge case where no changes were detected and so the build marks as a failure
https://github.com/hyperledger/caliper/actions/runs/2766550945

davidkel (approved):
LGTM

PR #1424, opened 2022-08-01
Add GitHub Actions workflow for docs

CaptainIRS:
In this PR:
* GitHub actions workflows are added to the gh-pages branch
* Issues pointed out by DCI lint are fixed for the latest and vNext docs

This is done because according to [GitHub Docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow):
> The following steps occur to trigger a workflow run:
> ...
> 2. GitHub searches the .github/workflows directory in your repository for workflow files that are present **in the associated commit SHA or Git ref of the event**.

So, the workflow file is expected to be present in the gh-pages branch for it to run.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,4 +1,4 @@
-name: CI (PR)
+name: Pages CI (PR)
 
 on:
   pull_request:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,9 @@
+name: CI (PR)
+
+on:
+  pull_request:
+    branches: [ "gh-pages" ]
+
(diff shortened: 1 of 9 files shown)


.github/workflows/pr.yml
```diff
@@ -0,0 +1,9 @@
+name: CI (PR)
```
davidkel, line 1: could we give this a different workflow name. There is an action called `pages-build-deployment` which is run when a gh-pages PR is merged so maybe we could call it `pages CI PR` ?

davidkel (changes requested):
one minor change on the workflow name, but looks good

davidkel (approved):
LGTM

PR #1427, opened 2022-08-04
Distinguish different workers in Prometheus PushGateway

### LINKED ISSUE #1354: Prometheus push gateway can't distinguish different workers and other issues
Summary of issues with the Prometheus Push Gateway (Also likely to affect the prometheus scrape option as well)
- Can't distinguish between different workers

### Can't distinguish between different workers
If I run say 5 workers, the push gateway only tracks the worker that sent the most recent update so when metrics are scraped you get a single entry for one of the workers, eg
when I query I get a single entry for a random worker
```
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="1"} 2148
```
whereas I would expect there to be 5 entries for each worker eg
```
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="0"} 1094
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="1"} 1035
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="2"} 1135
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="3"} 1157
caliper_tx_submitted{instance="",job="workers",roundIndex="0",roundLabel="empty-contract-evaluate",workerIndex="4"} 1091
```

removed info about txfinished, looks like it does reset ok

CaptainIRS:
In this PR:
* The worker index of each worker is added as a label when pushing metrics to PushGateway to distinguish workers.

Fixes #1354 

#### /metrics endpoint of PushGateway
image

#### Grafana dashboard showing metrics from two workers simultaneously
image

### DIFF
--- a/packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js
+++ b/packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js
@@ -111,7 +111,9 @@ class PrometheusPushTxObserver extends TxObserverInterface {
      * @private
      */
     async _sendUpdate() {
-        this.prometheusPushGateway.pushAdd({jobName: 'workers'}, function(err, _resp, _body) {
+        this.prometheusPushGateway.pushAdd({jobName: 'workers', groupings: {
+            'workerIndex': this.workerIndex,
+        }}, function(err, _resp, _body) {
             if (err) {
                 Logger.error(`Error sending update to Prometheus Push Gateway: ${err.stack}`);
             }
--- a/packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js
+++ b/packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js
@@ -45,8 +45,6 @@ class PrometheusPushTxObserver extends TxObserverInterface {
 
         // automatically apply default internal and user supplied labels
         this.defaultLabels = (options && options.defaultLabels) ? options.defaultLabels : {};
-        this.defaultLabels.workerIndex = this.workerIndex;
-        this.defaultLabels.roundIndex = this.currentRound;
         this.defaultLabels.roundLabel = this.roundLabel;
         this.registry.setDefaultLabels(this.defaultLabels);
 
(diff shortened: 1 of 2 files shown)


packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js
```diff
@@ -111,7 +111,9 @@ class PrometheusPushTxObserver extends TxObserverInterface {
      * @private
      */
     async _sendUpdate() {
-        this.prometheusPushGateway.pushAdd({jobName: 'workers'}, function(err, _resp, _body) {
+        this.prometheusPushGateway.pushAdd({jobName: 'workers', groupings: {
+            'workerIndex': this.workerIndex,
```
davidkel, line 115: We should also include in the groupings the round number so we can distinguish rounds and I think the job name should be more explicit
```suggestion
this.prometheusPushGateway.pushAdd({jobName: 'caliper-worker', groupings: {workerIndex: this.workerIndex, roundIndex: this.currentRound }}
```

We should also remove these lines in this file as well
```javascript
        this.defaultLabels.workerIndex = this.workerIndex;
        this.defaultLabels.roundIndex = this.currentRound;
```
In the constructor and activate as we now have groupings for them, this will require some minor updates to the unit tests

davidkel (changes requested):
With every code change in caliper we need to add unit tests, and so we need a new unit test for this change as I don't think there is a unit test that covers this code

CaptainIRS (commented):
I have looked at https://github.com/hyperledger/caliper/blob/main/packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js, but there doesn't seem to be any tests covering `_sendUpdate` as it is being stubbed out in unit tests. I'm not sure how a unit test for this would look like as pushAdd is a method from the `prom-client` library. Should the unit test just ensure that `workerIndex` is being passed to the `pushAdd` method (by using mocks)?

davidkel:
We would need a stub of the method `this.prometheusPushGateway.pushAdd` so we can track that it is being called and passing the expected parameters. This will then give us code coverage of that line and we know that it executes and passes the correct parameters, also checks that the call is likely to succeed and not fail when executed for real.

davidkel (approved):
LGTM

PR #1428, opened 2022-08-05
Add coverage report workflow

### LINKED ISSUE #1278: Code coverage reporting integrated into the build
### Please share the technical limitation of Caliper that you encountered.

No code coverage report is available for PRs

### Please detail your feature idea that could alleviate the limitation.

IT would be good for the build to produce a code coverage report. This would allow maintainers to review the PR more easily wereas at the moment a maintainer has to checkout the PR and run the tests to get a report. Also it could be useful for submitters as well to review.

### Please share some details about your use case if possible, and how the new feature would make Caliper a better performance benchmarking framework.

_No response_

### Please share any suggestions about the new feature's code/configuration API (using formatted YAML segments or pseudo-code).

_No response_

CaptainIRS:
In this PR:
* A step which reports coverage to Codecov is added to the unit tests workflow

Fixes #1278

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,24 @@
+name: Coverage Reports
+
+on:
+  workflow_call:
+
+jobs:
+  coverage:
+    name: Coverage Reports
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package: [caliper-core, caliper-fabric, generator-caliper]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Download coverage reports artifact
+      uses: actions/download-artifact@v1
+      with:
+        name: coverage-reports
+        path: packages
+    - name: Upload coverage reports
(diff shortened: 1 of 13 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  workflow_call:
+
+jobs:
+  coverage:
+    name: Coverage Reports
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package: [caliper-core, caliper-fabric, generator-caliper]
```
davidkel, line 12: do we not want all the other packages ?

CaptainIRS: This is because we're running unit tests only for these packages (since only these packages have unit tests). Most other packages simply run the linter when `npm test` is run, so we wouldn't have coverage information.

davidkel (approved):
LGTM

PR #1435, opened 2022-08-06
Terminate Caliper workers on interrupting Caliper manager

### LINKED ISSUE #1361: User termination of caliper should also terminate engaged remote workers
If a user for example decides to ctrl-c the caliper  manager, the remote workers it's engaged continue to work. Caliper Manager should send termination messages to these workers when caliper manager terminates regardless of why it terminated.

CaptainIRS:
In this PR:
* The SIGINT signal is handled in the manager to cleanly terminate workers when user tries to terminate Caliper manager.

Fixes #1361

### DIFF
--- a/packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
+++ b/packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
@@ -177,12 +177,6 @@ class RoundOrchestrator {
             logger.error(`Could not prepare worker connections: ${err.stack || err}`);
         }
 
-        process.on('SIGINT', async () => {
-            logger.info('SIGINT received, stopping workers');
-            await this._cleanup();
-            process.exit(0);
-        });
-
         let benchStartTime = Date.now();
 
         for (const [index, roundConfig] of roundConfigs.entries()) {
@@ -238,7 +232,7 @@ class RoundOrchestrator {
             logger.error(`Error while finalizing the report: ${err.stack || err}`);
         }
 
-        await this._cleanup();
+        await this.stop();
 
         let benchEndTime = Date.now();
         logger.info(`Benchmark finished in ${(benchEndTime - benchStartTime)/1000.0} seconds. Total rounds: ${success + failed}. Successful rounds: ${success}. Failed rounds: ${failed}.`);
@@ -247,7 +241,7 @@ class RoundOrchestrator {
     /**
      * Stops the benchmark.
      */
-    async _cleanup() {
+    async stop() {
         // clean up, with "silent" failure handling
         try {
             await this.monitorOrchestrator.stopAllMonitors();
--- a/packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
+++ b/packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
@@ -177,6 +177,12 @@ class RoundOrchestrator {
             logger.error(`Could not prepare worker connections: ${err.stack || err}`);
         }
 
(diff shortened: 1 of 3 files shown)


packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
```diff
@@ -177,6 +177,12 @@ class RoundOrchestrator {
             logger.error(`Could not prepare worker connections: ${err.stack || err}`);
         }
 
+        process.on('SIGINT', async () => {
```
davidkel, line 180: I don't think this should be registered here. Caliper has the separation of concerns as follows
launchManager (CLI) --> caliper-engine (Core) --> round-orchestrator (Core). I think this hook should be registered in the CLI layer and then call a `stop` api on caliper-engine that in turn calls stop on the round-orchestrator. 

Caliper is an api and can be embedded in different ways the launchManager is an example of how to embed the caliper engine so I think ownership of handling signals should be done by the embedding code rather than lower down. 
Also I think it's good to expose stop methods at this level. It allows embedding code to be able to stop based on whatever reason they see fit and caliper itself then has a method point to add further code to cleanly stop on request although at the moment I think it will just be 
launchManager:
```javascript
            process.on('SIGINT', async () => {
                // support a force stop
                if (!this.terminationAlreadyRequested) {
                    this.terminationAlreadyRequested = true;
                    logger.info('request to terminate received, stopping workers');
                    try {
                        await engine.stop();
                    } catch(error) {
                        logger.error(`failed to stop workers: ${error}`);
                        process.exit(1);
                    }
                    logger.info('Workers terminated.....');
                    process.exit(0);
                } else {
                    logger.info('Termination already requested, terminating immediately');
                    process.exit(2);
                }
            });
```
caliper-engine (will need to keep a reference to round orchestrator now)
```javascript
    async stop() {
        if (this.roundOrchestrator) {
            await this.roundOrchestrator.stop();
        }
    }
```

packages/caliper-core/lib/manager/orchestrators/round-orchestrator.js
```diff
@@ -225,14 +231,24 @@ class RoundOrchestrator {
...
+        await this._cleanup();
+
+        let benchEndTime = Date.now();
+        logger.info(`Benchmark finished in ${(benchEndTime - benchStartTime)/1000.0} seconds. Total rounds: ${success + failed}. Successful rounds: ${success}. Failed rounds: ${failed}.`);
+    }
+
+    /**
+     * Stops the benchmark.
+     */
+    async _cleanup() {
```
davidkel, line 250: We should call this `stop` for caliper-engine to call, but it must also call `this.workerOrchestrator.stop` otherwise it won't send the exit message to workers which is what should terminate the workers. 
eg
```suggestion
    async stop() {
        try {
            await this.monitorOrchestrator.stopAllMonitors();
        } catch (err) {
            logger.error(`Error while stopping monitors: ${err.stack || err}`);
        }

        try {
            await this.workerOrchestrator.stop();
        } catch (err) {
            logger.error(`Error while stopping workers: ${err.stack || err}`);
        }
    }
```

davidkel (changes requested):
Unfortunately this doesn't solve the problem, we need to ensure all workers receive the exit message to terminate them

davidkel (approved):
Even though there are no tests and I don't believe the submitter is going to return to update or submit any more PRs I think this is a good capability to have. So approving

davidkel:
Closing as DCO is proving to be an issue. Replicated this PR in #1514

Issue #1353, opened 2022-05-24
Providing Caliper Metrics to Prometheus


davidkel:
### Please share the technical limitation of Caliper that you encountered.

x

### Please detail your feature idea that could alleviate the limitation.

x

davidkel:
When we move the scrape mechanism from worker to manager we will lose the ability to scrape system metrics for individual workers which the prometheus client provides as it exposes them directly via prometheus. We would want to still capture the same information and forward this worker information back to the manager to collate and scrape so maybe #1043 can help with capturing those metrics so we can forward them back

PR #1448, opened 2022-09-21
Add TxObserver for Prometheus manager

CaptainIRS:
In this PR:
* The new TxObserver class is added (with tests) as a part of the ongoing effort to add a Prometheus scrape server to the Caliper manager.

PR 1/n for #1353

### DIFF
--- a/packages/caliper-core/lib/common/messages/prometheusUpdateMessage.js
+++ b/packages/caliper-core/lib/common/messages/prometheusUpdateMessage.js
@@ -1,0 +1,37 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const Message = require('./message');
+const MessageTypes = require('./../utils/constants').Messages.Types;
+
+/**
+ * Class for the "prometheusUpdate" message.
+ */
+class PrometheusUpdateMessage extends Message {
+    /**
+     * Constructor for a "prometheusUpdate" message instance.
+     * @param {string} sender The sender of the message.
+     * @param {string[]} recipients The recipients of the message.
+     * @param {object} content The content of the message.
+     * @param {string} date The date string of the message.
+     * @param {string} error The potential error associated with the message.
+     */
+    constructor(sender, recipients, content, date = undefined, error = undefined) {
+        super(sender, recipients, MessageTypes.PrometheusUpdate, content, date, error);
+    }
+}
+
+module.exports = PrometheusUpdateMessage;
--- a/packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
+++ b/packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
@@ -36,29 +36,42 @@ class PrometheusManagerTxObserver extends TxObserverInterface {
         super(messenger, workerIndex);
 
         this.method = (options && options.method) ? options.method : ConfigUtil.get(ConfigUtil.keys.Observer.PrometheusManager.Method);
-        if (this.method === 'periodic') {
+
+        switch (this.method) {
+        case 'periodic': {
             this.updateInterval = (options && options.interval) ? options.interval : ConfigUtil.get(ConfigUtil.keys.Observer.PrometheusManager.Interval);
             this.intervalObject = undefined;
             if (this.updateInterval <= 0) {
-                Logger.error('Invalid update interval specified, using default value');
                 this.updateInterval = ConfigUtil.get(ConfigUtil.keys.Observer.PrometheusManager.Interval);
+                Logger.warn(`Invalid update interval specified, using default value of ${this.updateInterval}`);
             }
             if (options && options.collationCount) {
                 Logger.warn('Collation count is ignored when using periodic method');
             }
-        } else if (this.method === 'collate') {
+            break;
+        }
+
+        case 'collate' : {
             this.collationCount = (options && options.collationCount) ? options.collationCount : ConfigUtil.get(ConfigUtil.keys.Observer.PrometheusManager.CollationCount);
             if (this.collationCount <= 0) {
-                Logger.error('Invalid collation count specified, using default value');
                 this.collationCount = ConfigUtil.get(ConfigUtil.keys.Observer.PrometheusManager.CollationCount);
+                Logger.warn(`Invalid collation count specified, using default value of ${this.collationCount}`);
             }
             if (options && options.interval) {
                 Logger.warn('Update interval is ignored when using collate method');
             }
+            break;
+        }
+
+        default: {
+            const msg = `Unrecognised method '${this.method}' specified for prometheus manager, must be either 'collate' or 'periodic' `;
+            Logger.error(msg);
+            throw new Error(msg);
+        }
+
         }
 
         this.pendingMessages = [];
-
         this.managerUuid = managerUuid;
     }
     /**
@@ -117,7 +130,7 @@ class PrometheusManagerTxObserver extends TxObserverInterface {
         this.pendingMessages.push(message);
 
         if (this.method === 'collate' && this.pendingMessages.length === this.collationCount) {
-            await this._sendUpdate();
+            this._sendUpdate();
         }
     }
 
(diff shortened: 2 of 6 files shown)


packages/caliper-core/lib/common/messages/prometheusUpdateMessage.js
```diff
@@ -0,0 +1,37 @@
...
+
+'use strict';
+
+const Message = require('./message');
+const MessageTypes = require('./../utils/constants').Messages.Types;
+
+/**
+ * Class for the "prometheusUpdate" message.
+ */
+class PrometheusUpdateMessage extends Message {
```
davidkel, line 23: I think we should call this a WorkerMetricsMessage as it isn't specific to prometheus here

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
```diff
@@ -0,0 +1,97 @@
+/*
```
davidkel, line 1: This isn't prometheus specific at this point so I think we could just call this ManagerTxObserver and the file manager-tx-observer,js

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
```diff
@@ -0,0 +1,97 @@
...
+            }
+        } else {
+            // pass/fail status from result.GetStatus()
+            const message = new PrometheusUpdateMessage(this.messenger.getUUID(), [this.managerUuid], {
+                event: 'txFinished',
+                workerIndex: this.workerIndex,
+                roundIndex: this.currentRound,
+                roundLabel: this.roundLabel,
+                status: results.GetStatus(),
+                latency: (results.GetTimeFinal() - results.GetTimeCreate())/1000
```
davidkel, line 78: Here you divide latency by 1000, but above you don't. I assume they should both be the same

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
```diff
@@ -0,0 +1,97 @@
...
+    constructor(options, messenger, workerIndex, managerUuid) {
+        super(messenger, workerIndex);
+
+        this.managerUuid = managerUuid;
+    }
+    /**
+     * Called when TXs are submitted.
+     * @param {number} count The number of submitted TXs. Can be greater than one for a batch of TXs.
+     */
+    txSubmitted(count) {
```
davidkel, line 41: I see that we send messages for every submit and finished txn, this could result in a large number of messages being sent on high tps runs. This may be fine but it's also not tuneable if there is an issue. It would be good to allow for periodic sending as well with the workers collating the information so that we can do bulk sending.

packages/caliper-core/test/worker/tx-observers/prometheus-manager-tx-observer.js
```diff
@@ -0,0 +1,214 @@
...
+     * @param {*} path path
+     * @return {string} the fake path
+     */
+    static resolvePath(path) {
+        return 'fake/path';
+    }
+
+    /**
+     *
+     * @return {boolean} the fake path
```
davidkel, line 38: comment doesn't match code

packages/caliper-core/test/worker/tx-observers/prometheus-manager-tx-observer.js
```diff
@@ -0,0 +1,214 @@
...
+    static augmentUrlWithBasicAuth(url) {
+        return url;
+    }
+}
+
+mockery.enable({
+    warnOnReplace: false,
+    warnOnUnregistered: false
+});
+mockery.registerMock('../../common/utils/caliper-utils', Utils);
```
davidkel, line 76: Is this actually needed ? the class under test doesn't seem to use the Utils package

davidkel (changes requested):
Good start, some comments about naming and also would like to see the option for periodic bulk messaging

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
CaptainIRS: I have named it this way because it would be inline with the transaction monitor module in the config file. In the existing codebase `module: prometheus-push` is bound to a `PrometheusPushTxObserver` and `module: prometheus` is bound to a `PrometheusTxObserver`, etc., so `PrometheusManagerTxObserver` would be the equivalent for the planned `module: prometheus-manager` config property: https://github.com/hyperledger/caliper/blob/4aa36ea91e21966d4aac65d04a36720ee93ca36d/packages/caliper-core/lib/worker/tx-observers/tx-observer-dispatch.js#L21-L25

Should I perform the refactor to `ManagerTxObserver`?

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
CaptainIRS: I had noticed this, but I had left it as such so that it is inline with the implementations in the other prometheus observers:

https://github.com/hyperledger/caliper/blob/5938426a10af642989da0bb4fceeeb185e24e4aa/packages/caliper-core/lib/worker/tx-observers/prometheus-tx-observer.js#L163-L175

https://github.com/hyperledger/caliper/blob/c648fa9510fdb856d9cfec455dc8eb42cee2600b/packages/caliper-core/lib/worker/tx-observers/prometheus-push-tx-observer.js#L165-L177

I would like to know if this is a bug in the codebase or is it intended. If this is a bug, I could fix this for the other observers in a separate PR.

packages/caliper-core/test/worker/tx-observers/prometheus-manager-tx-observer.js
CaptainIRS: Note to self: This needs to be fixed in https://github.com/hyperledger/caliper/blob/c648fa9510fdb856d9cfec455dc8eb42cee2600b/packages/caliper-core/test/worker/tx-observers/prometheus-push-tx-observer.js#L39 and https://github.com/hyperledger/caliper/blob/04a1154fcbc530131b3c03fa97a7ecb833660f56/packages/caliper-core/test/worker/tx-observers/prometheus-tx-observer.js#L39 as well in a later PR

packages/caliper-core/test/worker/tx-observers/prometheus-manager-tx-observer.js
CaptainIRS: This is needed because the base interface (TxObserverInterface) is dependent on the Utils class. Without the mock, this error would be thrown on running the tests:
```  1) When using a PrometheusManagerTxObserver should set managerUuid passed through constructor:
     Error: Util.resolvePath: Parameter is undefined
      at Function.resolvePath (lib/common/utils/caliper-utils.js:247:19)
      at new TxObserverInterface (lib/worker/tx-observers/tx-observer-interface.js:26:61)
      at new PrometheusManagerTxObserver (lib/worker/tx-observers/prometheus-manager-tx-observer.js:21:105)
      at new createTxObserver (lib/worker/tx-observers/prometheus-manager-tx-observer.js:36:123)
      at Context.<anonymous> (test/worker/tx-observers/prometheus-manager-tx-observer.js:90:26)
      at processImmediate (node:internal/timers:466:21)```

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
CaptainIRS: I have a few doubts regarding how this could be implemented.

I'm thinking of the following approach:
1. For the periodic-sending feature: A timer is started during the instantiation of the observer, which would regularly send the messages present in an array that keeps track of the pending messages.
2. For the collation feature: The max collation size is obtained in the config and an array is maintained with pending messages. When a new message arrives and the collation size exceeds on appending the message, the pending messages are sent in bulk.

However I'm not sure about how we could handle the sending of messages during termination of Caliper after the benchmark is successful. In both these features, there might be some loss of data as:
1. In case of periodic sending, the timer for the last batch of messages might not be sent as the timer might trigger after the scrape server is shut down during the termination of Caliper.
2. In case of collation, the last batch of messages might have a count less than the size defined in the config, in which case, it would not be sent when Caliper terminates.

I would like to know if the approaches would be feasible and if there is any way we could handle these cases.

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
davidkel: That looks like a bug to me in the other observers and both entries should divide by 1000 as it looks like the latency histogram is specified as seconds

packages/caliper-core/lib/worker/tx-observers/prometheus-manager-tx-observer.js
davidkel: As discussed the deactivate method should help in handling the termination of a worker case

davidkel (changes requested):
Left details on discord

davidkel (approved):
LGTM

PR #1456, opened 2022-10-10
Fix calculation and doc comment in prometheus observer

CaptainIRS:
* Fixed histogram calculation
* Fixed doc comment in test utility class


davidkel (approved):
LGTM

PR #1458, opened 2022-10-17
Fix integration test CI trigger

### LINKED ISSUE #1441: Should run integration tests if a change to package-lock.json PR is submitted by dependabot
We probably should run all the integration tests if dependabot submits an update to package-lock.json as we don't know easily which components it affects

CaptainIRS:
* Run integration tests when package-lock.json or the workflow file are modified

Fixes #1441


davidkel (approved):
LGTM

PR #1460, opened 2022-10-17
Revert "Add Node 18 to CI (#1455)"

CaptainIRS:
The fisco-bcos adapter and the generator don't seem to be compatible with node 18, hence the change is reverted until it is fixed.


davidkel (approved):
LGTM

PR #1465, opened 2022-10-26
[WIP] Experimental integration test runner using microfab

CaptainIRS:
Some issues/notes as of now:
* TLS (see IBM-Blockchain/microfab#105)
* ~~There seems to be an issue selecting the correct endorsing peers when using the older Fabric SDK (works well using the newer Fabric Gateway API though)~~
* ~~Normal installation of chaincode doesn't seem to work in microfab (can only install using CCAS)~~

### DIFF
--- a/packages/caliper-tests-integration/fabric_tests/.gitignore
+++ b/packages/caliper-tests-integration/fabric_tests/.gitignore
@@ -1,4 +1,9 @@
 **/*.log
+*.tar.gz
+_cfg
+bin
+config
 report.html
 fabric-samples
 vendor
+microfab
--- a/packages/caliper-tests-integration/fabric_tests/docker-compose.yaml
+++ b/packages/caliper-tests-integration/fabric_tests/docker-compose.yaml
@@ -51,7 +51,7 @@ services:
         - /sys:/sys:ro
         - /var/lib/docker/:/var/lib/docker:ro
         ports:
-        - 8080:8080
+        - 8888:8080
         restart: always
 
 ###############
@@ -68,8 +68,3 @@ services:
         - "9001:9001"
         volumes:
         - ./mosquitto/mosquitto.conf:/mosquitto/config/mosquitto.conf
-
-networks:
-  default:
-    name: fabric_test
-    external: true
--- a/packages/caliper-tests-integration/fabric_tests/init.js
+++ b/packages/caliper-tests-integration/fabric_tests/init.js
@@ -64,7 +64,7 @@ class MarblesInitWorkload extends WorkloadModuleBase {
             contractId: this.txIndex % 2 === 0 ? 'mymarbles' : 'yourmarbles',
             contractFunction: 'initMarble',
             contractArguments: [marbleName, marbleColor, marbleSize, marbleOwner],
-            invokerIdentity: 'client0.org1.example.com',
+            invokerIdentity: 'client0.org1.127-0-0-1.nip.io:8080',
             timeout: 5,
             readOnly: false
         };
--- a/packages/caliper-tests-integration/fabric_tests/initByChannel.js
+++ b/packages/caliper-tests-integration/fabric_tests/initByChannel.js
@@ -64,10 +64,10 @@ class MarblesInitByChannelWorkload extends WorkloadModuleBase {
             contractId: this.txIndex % 2 === 0 ? 'mymarbles' : 'yourmarbles',
             contractVersion: 'v0',
             contractFunction: 'initMarble',
-            invokerIdentity: 'client0.org1.example.com',
+            invokerIdentity: 'client0.org1.127-0-0-1.nip.io:8080',
             channel: this.txIndex % 2 === 0 ? 'mychannel' : 'yourchannel',
             contractArguments: [marbleName, marbleColor, marbleSize, marbleOwner],
-            timeout: 5,
+            timeout: 10,
             readOnly: false
         };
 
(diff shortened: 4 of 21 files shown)


CaptainIRS (commented):
@davidkel 

> Can you elaborate on selecting the correct endorsing peers.

The following error occurs randomly and can be observed in the [GitHub Action logs](https://github.com/hyperledger/caliper/actions/runs/3331701902/jobs/5511781438#step:9:941):
```log
2022-10-26T18:56:56.449Z - error: [TransactionEventStrategy]: constructor: No peers for strategy
2022.10.26-18:56:56.450 +0000 error [connectors/v2/FabricGateway] Failed to perform submit transaction [initMarble] using arguments [marbles_phase_4_0_0_0,red,10,Alice],  with error: Error: No peers for strategy
    at new TransactionEventStrategy (/tmp/sut/cached/v2.2.14/node_modules/fabric-network/lib/impl/event/transactioneventstrategy.js:49:19)
    at new AnyForTxStrategy (/tmp/sut/cached/v2.2.14/node_modules/fabric-network/lib/impl/event/anyfortxstrategy.js:41:1)
    at Transaction.MSPID_SCOPE_ANYFORTX [as eventHandlerStrategyFactory] (/tmp/sut/cached/v2.2.14/node_modules/fabric-network/lib/impl/event/defaulteventhandlerstrategies.js:60:27)
    at Transaction.submit (/tmp/sut/cached/v2.2.14/node_modules/fabric-network/lib/transaction.js:257:39)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async V2FabricGateway._submitOrEvaluateTransaction (/home/runner/work/caliper/caliper/packages/caliper-fabric/lib/connector-versions/v2/FabricGateway.js:375:26)
    at async V2FabricGateway._sendSingleRequest (/home/runner/work/caliper/caliper/packages/caliper-fabric/lib/connector-versions/v2/FabricGateway.js:168:16)
    at async V2FabricGateway.sendRequests (/home/runner/work/caliper/caliper/packages/caliper-core/lib/common/core/connector-base.js:83:26)
    at async MarblesInitWorkload.submitTransaction (/home/runner/work/caliper/caliper/packages/caliper-tests-integration/fabric_tests/init.js:72:9)
```

On trying to debug by setting a breakpoint:

image

Here, `network.channel.endorsers` has only one peer, belonging to Org2, whereas the requested `mspId` is of Org1, thus leading to no endorsing peers. This however happens only with SDKs <= v2.2 and doesn't appear with the Fabric Gateway APIs in v2.4. I'm not really sure about the root cause of this issue though.

> we shouldn't need to resort to CCAAS can you provide some more info on what you saw

Sorry, I must've had some incorrect configuration when I tested earlier, but now it works and I have updated the PR to use the normal chaincode packaging method.

davidkel:
@CaptainIRS the errors are very odd and the fact they happen in both 1.4 and 2.2 node sdks only with microfab. Not surprised with the fabric-gateway as it does event handing differently. I'll see if I can get someone to look at it too.

mbwhite:
Hello @CaptainIRS ...  I've been trying to recreate this problem with Microfab, without much success I'm afraid.
I believe that (logically) there are 2 processes, each running an application sending transactions. Is each just doing one org are does it swap?

I've got two apps here that are using the 2.2 NodeSDK submitting transactions to microfab; it seems completely stable.

Chaincode Signature Policy is `--signature-policy "OR('org1MSP.peer','org2MSP.peer')"`
Microfab config ```    export MICROFAB_CONFIG='{
        "endorsing_organizations":[
            {
                "name": "org1"
            },
            {
                "name": "org2"
            }
        ],
        "channels":[
            {
                "name": "mychannel",
                "endorsing_organizations":[
                    "org1","org2"
                ]
            }

        ],
        "capability_level":"V2_0"
    }'
```

**Review by CaptainIRS — COMMENTED** [review]
Hi @mbwhite I have tried using just one channel in CaptainIRS/caliper#3, but the same error seems to occur even in this case. (GitHub Action logs: [here](https://github.com/CaptainIRS/caliper/actions/runs/3339244954/jobs/5527980283#step:9:1001) and [here](https://github.com/CaptainIRS/caliper/actions/runs/3339244954/jobs/5527980409#step:9:1231))

**Comment by mbwhite** [comment]
@CaptainIRS found the problem!!

For microfab the `asLocalhost` setting _must_ be false.  Irrespective of the fact it is local... it needs the full hostnames.

```
caliper:
    fabric:
       gateway:
         localhost: false
```

Added that to the my local caliper configuration `caliper.yaml` and all is well.

**Comment by davidkel** [comment]
Thank you @mbwhite for investigating. @CaptainIRS Best way is to add
```
--caliper-fabric-gateway-localhost false
```
when launching caliper eg
```
${CALL_METHOD} launch manager --caliper-workspace phase4 --caliper-flow-only-test --caliper-fabric-gateway-localhost false
```

It's interesting that microfab appears to work using the legacy sdks when localhost is set to true for a lot of the time so definitely something that microfab should document. I thought it used hostname for routing and as the endpoint urls get translated to localhost:8080 I wasn't expecting that to work at all. The problem was that the endorsers were being discovered and the endpoints were being translated to the same endpoint of localhost:8080 from org1peer-api.127-0-0-1.nip.io:8080 and org2peer-api.127-0-0-1.nip.io:8080 and thus you could only have one at that endpoint so in some cases was being loaded with the wrong peer for the mspid of the identity. I also suspect that if the chaincode had an AND endorsement policy it would have failed for all submissions as it wouldn't be able to get all endorsements as there would only be 1 endorser in the channel object.

CaptainIRS (commented):
It works after adding the CLI flag for localhost=false. Thanks @mbwhite for the help.

> I also suspect that if the chaincode had an AND endorsement policy it would have failed for all submissions

@davidkel, I have tested this earlier and it indeed fails for all submissions.

davidkel:
Closing this pull request because microfab is has gone dormant with the maintainers now not involved in hyperledger fabric anymore and it's unlikely to be maintained in the future.

PR #1484, opened 2023-02-21
Use go chaincode in integration tests

### LINKED ISSUE #1476: remove node as the chaincode language for fabric integration testing
Couple of reasons for this
1. Fabric doesn't pin exact versions of it's dependencies which means that changes to the dependency tree outside of our control (in fact it's outside of fabric's control because they don't pin dependencies) can break the code. This has happened today with grpc-js 1.8.1->1.8.2 breaking all node chaincodes. We could add a shrinkwrap to fix the versions but that brings me to my second point
2. Node chaincode is not as efficient as Go chaincode. Also npm install has to be run to prepare the chaincode for each run whereas we go chaincode we can pre-vendor and keep that in the code tree. Both will make it faster for integration testing and reduce the testing time and thus reduce power usage. Also this is less likely to break.

CaptainIRS:
In this PR:
* Node chaincode is removed from integration tests
* All integration tests now use go chaincode
* Go chaincode is updated to the latest version in the main branch of fabric-samples

Fixes #1476


davidkel (approved):
LGTM

PR #1506, opened 2023-10-24
Resend tx when it fails on ethereum

NickSneo:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,10 @@
             "scripts",
             ".git",
             ".gitignore",
+            ".npmrc",
             "packages",
-            "log"
+            "log",
+            "dev-docs/figures/.gitkeep"
         ],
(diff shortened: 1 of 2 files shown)


davidkel (changes requested):
I'm not sure what this will achieve, caliper is designed to track success and failure of individual transations so why mark it as a failure then retry ? Also there is the potential for this to resend repeatedly as there is no logic to limit the amount of resend sent before failing completely. 

If you could raise an isssue to discuss the particular issue you are seeing and the benefit/reasoning behind resending the transaction on a failure (rather than having workloads handle this for example) then it it can be moved forward but as it stands I don't think this change is appropriate.

davidkel:
Closing as no response from submitter

PR #1514, opened 2024-02-21
Terminate workers if caliper manager is terminated prematurely

davidkel:
Caliper can run workers in multiple ways. THe most common is when they are part of the Caliper manager process, however when they are separate processes either on the same machine or remote machines, if the caliper manager is terminated unexpectedly (eg CTRL-C) then those workers still continue to drive load. We would want those workers to terminate as well. This change facilitates this by detecting if the manager is terminated prematurely and sending requests to the workers to terminate

This is a resubmitted PR of a collaboration between myself and @captainIRS (#1435 )

Signed-off-by: Dave Kelsey <d_kelsey@uk.ibm.com>


aklenik (approved):
LGTM

PR #1515, opened 2024-02-21
Remove fisco-bcos

davidkel:
Node 16 is now out of lts. Fisco bcos will not work on node 18 or higher due to it not being compatible with openssl 3 so caliper at the moment can't move it's builds to node 18 or node 20. Also openssl 1.1 is now out of support and so the connector is using crypto libraries which are no longer supported. As we have no-one who can maintain this connector, it should be dropped.


aklenik:
@davidkel I think we should identify components to be dropped, and publish a "dummy release" that states their deprecation. And then we should actually remove them in a subsequent version (giving the community ample time to react or protest). Opinions?

davidkel:
@aklenik unfortunately when it comes to fisco-bcos it's blocking us from moving our builds to node 18/20 as node 14/16 are not supported anymore (and thus no security fixes). People can still use caliper 0.5.0 if they want to use fisco-bcos of course. I've only seen 1 issue raised on github for fisco (it was a question and it went unanswered) and never any other questions on stackoverflow or discord. I don't even know if it works with the latest fisco-bcos releases and I would have to question the viability of a blockchain solution from an environment that can't even secure it's own website (see http://www.fisco-bcos.org/ there is no https:// version)

In principle for components though I agree, but in the case of fisco I think the risk is very low here that it would affect anyone and they could still use 0.5.0 (with node14/16) if they really wanted to

aklenik (approved):
I accept your argument. I'll drop an announcement on discord later about the removal.

PR #1519, opened 2024-03-25
Update the versions of fabric sdks

davidkel:
This rationalises the versions and we should only document the following binding versions

1, 1.4 (use 1.4 legacy sdk)
2.2 (use 2.2 legacy sdk)
2.4, 2.5, 3, fabric-gateway (use the fabric-gateway sdk)

We will keep 2.4 (for legacy reasons only as 2.4 was never LTS), 2.5 to cover the published versions of fabric we support however the binding of fabric-gateway should be the one going forward and seen in the documentation.
because Fabric 2 changed the recommended client version in a point release (2.5 deprecated the legacy sdks and the only real option if the fabric-gateway) otherwise we will specify a single digit version for fabric (1,3)

the 1.4, 2.2 sdks are now deprecated and 2.2 is now or version soon to be out of lts, therefore fabric-gateway is the only version that will and should be used in the future and there is no plan to change the client side much given the reduction in resource of people using it and the use of SmartBFT should not affect the client side version (I hope :-) )

### DIFF
--- a/packages/caliper-cli/lib/lib/config.yaml
+++ b/packages/caliper-cli/lib/lib/config.yaml
@@ -39,23 +39,18 @@ sut:
     # the 2.4 key without creating a new version and the integration tests will automatically pick it up.
     fabric:
         # The name/key of the SDK binding
-        1.4.11:
-            packages: ['fabric-client@1.4.11', 'fabric-network@1.4.11','fs-extra@8.1.0']
-        1.4.14:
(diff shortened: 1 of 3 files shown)


aklenik:
@davidkel there still seems to be some issue with the publishing, should we delay this pr until the publishing is fixed?

davidkel:
@aklenik I think so, hopefully my extra PR will get publishing working again for the time being until a proper fix can be done.

PR #1523, opened 2024-03-26
downgrade npm on publish as temp fix

davidkel:
downgrade the version of npm on the publish step to ensure that publish doesn't give an ENOWORKSPACES error

This should be considered a temporary solution to getting publish working again and a better solution needs to be found.


aklenik (approved):
LGTM

PR #1545, opened 2024-04-22
Remove collection of node gc stats in prometheus

davidkel:
reasons for this

1. we have never documented the metrics
2. not sure how much use it is anyway
3. was using an old version of gc-stats which is now unmaintained and has security issues
4. replacement version of gc-stats is maintained by a single person who controls the binary builds available so is a dangerous thing to rely on
5. it has no pure javascript fallback therefore requires caliper to pull down or build a binary module (meaning you need python and compilers installed) which will also be dependent on the version of node being used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,12 +27,12 @@
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.23.5",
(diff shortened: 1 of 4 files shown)


ryjones:
@davidkel I think you can merge now. If you can't, please let me know and I'll merge it

davidkel:
thanks @ryjones much appreciated

Issue #1557, opened 2024-04-24
Unit tests for caliper-core: Improve Coverage


davidkel:
current report shows
```
--------------------------------------------------------------|----------|----------|----------|----------|----------------|
File                                                          |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------------------------------------------------------|----------|----------|----------|----------|----------------|
All files                                                     |     48.7 |    39.91 |    41.18 |    48.74 |                |
 caliper-core                                                 |      100 |      100 |      100 |      100 |                |
  index.js                                                    |      100 |      100 |      100 |      100 |                |
 caliper-core/lib/common/config                               |     80.7 |    65.52 |      100 |     80.7 |                |
  Config.js                                                   |    76.09 |    58.33 |      100 |    76.09 |... 267,268,290 |
  config-util.js                                              |      100 |      100 |      100 |      100 |                |
 caliper-core/lib/common/core                                 |    82.81 |       90 |    71.88 |    81.77 |                |
  connector-base.js                                           |    86.54 |    83.33 |    55.56 |    86.54 |... 3,64,66,166 |
  connector-interface.js                                      |    18.18 |      100 |        0 |    18.18 |... 8,86,96,105 |
  transaction-statistics-collector.js                         |    96.15 |    93.75 |    98.21 |     95.7 |360,387,410,415 |
  transaction-status.js                                       |       48 |      100 |    40.91 |       48 |... 194,203,211 |
 caliper-core/lib/common/messages                             |    55.63 |     9.62 |    22.45 |    55.63 |                |
  assignIdMessage.js                                          |       60 |        0 |        0 |       60 |          33,41 |
  assignedMessage.js                                          |       75 |        0 |        0 |       75 |             32 |
  connectedMessage.js                                         |       75 |        0 |        0 |       75 |             32 |
  exitMessage.js                                              |       80 |        0 |        0 |       80 |             32 |
  initializeMessage.js                                        |       80 |        0 |        0 |       80 |             32 |
  message.js                                                  |    42.11 |     12.5 |    11.11 |    42.11 |... 140,141,144 |
  parse.js                                                    |    40.54 |        0 |        0 |    40.54 |... 72,74,76,78 |
  prepareMessage.js                                           |    21.43 |        0 |        0 |    21.43 |... ,98,106,114 |
  preparedMessage.js                                          |       75 |        0 |        0 |       75 |             32 |
[...]

PR #1561, opened 2024-05-01
`caliper-core`: add tests for `Message` classes

EshaanAgg:
This PR introduces the tests for the base `Message` class in `caliper-core`, and then introduces three similar test suites for the constructors of three types of messages. When approved, I can work on adding similar tests for all the message classes. 

## Checklist
 - [X]  A link to the issue/user story that the pull request relates to
 - [ ]  Design of the fix
 - [X]  Automated tests that prove the fix keeps on working

## Issue/User story
Fixes part of #1557 

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

### DIFF
--- a/packages/caliper-core/test/common/core/messages/message.js
+++ b/packages/caliper-core/test/common/core/messages/message.js
@@ -20,7 +20,7 @@ const AllMessageTarget = require('../../../../lib/common/utils/constants').Messa
 const chai = require('chai');
 chai.should();
 
-describe('Message', () => { 
+describe('Message', () => {
     const mockSender = 'Test User';
     const mockRecipients = ["recepient-id-1", "recepient-id-2", "recepient-id-3"];
     const mockType = 'Assigned';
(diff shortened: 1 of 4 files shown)


davidkel:
@aklenik Can you take a look

davidkel:
On a quick glance, these seem ok to me, but really want @aklenik to take a look as he is more familiar with this part of the code I think

packages/caliper-core/test/common/core/messages/message.js
```diff
@@ -0,0 +1,140 @@
...
+            message.hasError().should.be.true;
+        })
+
+        it("should return false if the message has no error", () => {
+            const message = new Message(mockSender, mockRecipients, mockType, mockContent, mockDate);
+            message.hasError().should.be.false;
+        })
+    })
+
+    describe("forRecipient", () => {
```
davidkel, line 93: We need tests to show that it returns false as well

davidkel (changes requested):
One comment around a couple more tests being required, but these seem reasonable to me.

packages/caliper-core/test/common/core/messages/message.js
EshaanAgg: My bad for missing this case! I have updated the test suite for the same. Thanks!

davidkel (approved):
Looks good to me, many thanks for the contribution

PR #1579, opened 2024-05-17
Implement part of #1557: Add tests for replayRate controller.

rahat2134:
## Checklist
 - [X]  A link to the issue/user story that the pull request relates to
 - [ ]  Design of the fix

## Issue/User story
Add tests for replayRate controller.

### DIFF
--- a/packages/caliper-core/test/worker/rate-control/replayRate.js
+++ b/packages/caliper-core/test/worker/rate-control/replayRate.js
@@ -1,0 +1,200 @@
+/*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const ReplayRate = require('../../../lib/worker/rate-control/replayRate');
+const TestMessage = require('../../../lib/common/messages/testMessage');
+const TransactionStatisticsCollector = require('../../../lib/common/core/transaction-statistics-collector');
+const caliper_utils = require('../../../lib/common/utils/caliper-utils');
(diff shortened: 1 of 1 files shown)


rahat2134:
@davidkel PTAL.

packages/caliper-core/test/worker/rate-control/replayRate.js
```diff
@@ -0,0 +1,200 @@
...
+        sandbox = sinon.createSandbox();
+        if (caliper_utils.sleep.isSinonProxy) {
+            sandbox.restore();
+        }
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
```
davidkel, line 42: we need to describe what was being done here as well, so we need a describe wrapping the first 3 it statements with a description such as 
`When creating the replay rate controller`

packages/caliper-core/test/worker/rate-control/replayRate.js
```diff
@@ -0,0 +1,200 @@
...
+        const stubStatsCollector = sinon.createStubInstance(TransactionStatisticsCollector);
+        const replayRateController = ReplayRate.createRateController(testMessage, stubStatsCollector, 0);
+
+        assert.deepEqual(replayRateController.records, transactionTimings);
+
+        // Clean up the temporary file
+        fs.unlinkSync(tempFilePath);
+    });
+
+    it('should correctly apply rate control and log warning when running out of transaction timings', async () => {
```
davidkel, line 114: There are 2 different tests here, can we split them into 2 please. One where there are no log warnings and one where there are log warnings

packages/caliper-core/test/worker/rate-control/replayRate.js
```diff
@@ -0,0 +1,200 @@
...
+        // Test the case when running out of transaction timings
+        await replayRateController.applyRateControl();
+        sinon.assert.calledWithExactly(sleepStub, 50);
+        sinon.assert.calledWithMatch(loggerWarnStub, /Using default sleep time of 50 ms/);
+
+        // Clean up the temporary file
+        fs.unlinkSync(tempFilePath);
+    });
+
+    it('should call end() method without any errors', async () => {
```
davidkel, line 172: We try to avoid method names in descriptions of tests and focus on behaviour descriptions. It's a way to understand what this test is trying to achieve. Then it also helps to understand what behaviours may also be missing in the tests. So for this one it is testing that the rate controller is never actually used, but what about when it is used ? We don't want be influenced by the implementation as it then keeps the tests viable for when changes are made to the implementation.

PR #1592, opened 2024-06-04
Add fixed-load rate controller to calliper generator

### LINKED ISSUE #1587: Caliper generator doesn't support the fixed-load rate controller
In index.js we see
```
    async writing () {
        console.log('Generating benchmark files...');
        this._workloadWrite();
        answersObject.rateController = promptAnswers.rateController;
        switch(promptAnswers.rateController) {
        case 'fixed-rate':
            answersObject.opts = 'tps: 10';
            this._configWrite();
            break;
        case 'fixed-backlog':
            answersObject.opts = 'unfinished_per_client: 5';
            this._configWrite();
            break;
        case 'linear-rate':
            answersObject.opts = 'startingTps: 25, finishingTps: 75';
            this._configWrite();
            break;
        case 'fixed-feedback-rate':
            answersObject.opts = 'tps: 100, unfinished_per_client: 100';
            this._configWrite();
            break;
        }
    }
```
fixed-load is not in there

duckling69:
fixes #1587

### DIFF
--- a/packages/generator-caliper/generators/benchmark/index.js
+++ b/packages/generator-caliper/generators/benchmark/index.js
@@ -272,6 +272,10 @@ module.exports = class extends Generator {
             answersObject.opts = 'tps: 100, unfinished_per_client: 100';
             this._configWrite();
             break;
+        case 'fixed-load':
+            answersObject.opts = 'startTps:10, transactionLoad:20';
+            this._configWrite();
+            break;
         }
     }
 


davidkel:
@duckling69 Hi thanks for the contribution, please could you add unit tests to please otherwise we can't accept the PR and as you can see the code coverage has reduced with the addition of the new code.

duckling69:
@davidkel did the review changes

packages/generator-caliper/generators/benchmark/index.js
```diff
@@ -272,6 +272,10 @@ module.exports = class extends Generator {
             answersObject.opts = 'tps: 100, unfinished_per_client: 100';
             this._configWrite();
             break;
+        case 'fixed-load':
+            answersObject.opts = 'startTps:10, transactionLoad:20';
```
davidkel, line 276: Could we have a single space added between property and value so that it's consistent with the others, ie
`startTps: 10, transactionLoad: 20`

davidkel (changes requested):
We should also include this as an option when running the yo generator. see around lines 101-106 of index.js to update

packages/generator-caliper/generators/benchmark/index.js
duckling69: done

packages/generator-caliper/generators/benchmark/index.js
```diff
@@ -104,6 +104,7 @@ module.exports = class extends Generator {
                     {name: 'Fixed Backlog', value: 'fixed-backlog'},
                     {name: 'Linear Rate', value: 'linear-rate'},
                     {name: 'Fixed Feedback Rate', value: 'fixed-feedback-rate'}
+                    {name: 'Fixed Load', value: "fixed-load"}
```
davidkel, line 107: Please run the tests locally before submitting a PR. Doing so will quickly catch bugs in your submission. Here the build fails because you put a bug into the code, you are missing a `,` at the end of line 106

davidkel (changes requested):
please fix the bug in index.js

davidkel (approved):
LGTM, many thanks for the contribution

Issue #1606, opened 2024-08-19
Increase some modules housed in the worker directory unit test coverage


tunedev:
This aims to increase the below-listed modules file path's test coverage

## Context

The modules are all under the caliper-core/lib/worker directory
Related to #1557

## Expected Behavior

After running the tests in caliper-core the below listed %stmts in the code coverage report should tally with the below listed
- caliper-core/lib/worker/caliper-worker.js: 82
- caliper-core/lib/worker/worker-message-handler.js: 9.78
- caliper-core/lib/worker/fixedFeedbackRate.js: 60
- caliper-core/lib/worker/recordRate.js: 43.28
![Screenshot from 2024-08-19 09-51-05](https://github.com/user-attachments/assets/75a32d33-739d-4cdf-8e64-08fa64c1495e)

## Actual Behavior

After running the tests in caliper-core the below listed %stmts in the code coverage report the below listed was what was there before
- caliper-core/lib/worker/caliper-worker.js: 93.75
- caliper-core/lib/worker/worker-message-handler.js: 89.13
- caliper-core/lib/worker/fixedFeedbackRate.js: 80
- caliper-core/lib/worker/recordRate.js: 95.52
![Screenshot from 2024-08-19 09-50-20](https://github.com/user-attachments/assets/8affc009-982f-4169-ba3a-282e9769e4d0)

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-caliper)
- [ ] [GitHub Issues](https://github.com/hyperledger/caliper/issues)
- [ ] [Discord Chat history](https://discord.com/channels/905194001349627914/941417677778473031)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

PR #1617, opened 2024-09-09
Add more test for tx observers

tunedev:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Partially addresses #1606

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Validation of the fix

After this PR: Running the tests in caliper-core the below listed %stmts in the code coverage report should tally with the below listed
- caliper-core/lib/worker/tx-observers: 93.06
- caliper-core/lib/worker/tx-observers/internal-tx-observer.js: 100
- caliper-core/lib/worker/tx-observers/logging-tx-observer.js: 100
- caliper-core/lib/worker/tx-observers/prometheus-tx-observer.js : 92.06
- caliper-core/lib/worker/tx-observers/tx-observer-dispatch.js : 92.59

Screenshot from 2024-09-09 13-36-44

Before the PR: Running the tests in caliper-core the below listed %stmts in the code coverage report the below listed was what was there before

- caliper-core/lib/worker/tx-observers: 68.82
- caliper-core/lib/worker/tx-observers/internal-tx-observer.js: 26.08
- caliper-core/lib/worker/tx-observers/logging-tx-observer.js: 0
- caliper-core/lib/worker/tx-observers/prometheus-tx-observer.js : 77.77
- caliper-core/lib/worker/tx-observers/tx-observer-dispatch.js : 19.23

Screenshot from 2024-09-09 13-40-12

### DIFF
--- a/packages/caliper-core/lib/worker/tx-observers/internal-tx-observer.js
+++ b/packages/caliper-core/lib/worker/tx-observers/internal-tx-observer.js
@@ -20,6 +20,8 @@ const TxResetMessage = require('../../common/messages/txResetMessage');
 const CaliperUtils = require('../../common/utils/caliper-utils');
 const ConfigUtil = require('../../common/config/config-util');
 
+const logger = CaliperUtils.getLogger('InternalTxObserver');
+
 /**
  * Internal TX observer used by the worker process to driver TX scheduling and report round statistics
  * It is always instantiated.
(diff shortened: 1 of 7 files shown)


packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +161,91 @@ describe('Caliper worker', () => {
             await worker.executeRound(mockTestMessage);
             validateCallsAndWarnings(4);
         });
+
+        it('should complete all transactions', async () => {
```
davidkel, line 165: This test isn't really testing the behaviour. Here you are testing the running a fixed number of transactions capability of the worker, so the description needs to match this, it needs to actually prove that it does run the fixed number of transactions and should cover at least 2 different runs ideally. The test should also state that no submit errors will occur so will wait for the transactions to finish.

davidkel (changes requested):
Unfortunately I think that there is too much in this PR. I think it would be easier to review if we have much smaller PRs and each PR should focus on a single unit (ie .js file under test). If we do it this way I think it will be easier to focus and also we can get code merged more quicky. Please could you split the PR into multiple ones

PR #1625, opened 2024-09-10
Add more tests for caliper worker

tunedev:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Partially addresses #1606

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Validation of the fix

- caliper-core/lib/worker/worker-message-handler.js: 89.13

- caliper-core/lib/worker/worker-message-handler.js: 9.78

### DIFF
--- a/packages/caliper-core/lib/worker/caliper-worker.js
+++ b/packages/caliper-core/lib/worker/caliper-worker.js
@@ -84,9 +84,10 @@ class CaliperWorker {
      * @param {Object} rateController rate controller object
      * @async
      */
-    async runFixedNumber(workloadModule, number, rateController) {
+    async _runFixedNumber(workloadModule, number, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
         let error = undefined;
+
         while (stats.getTotalSubmittedTx() < number && !error) {
             await rateController.applyRateControl();
 
@@ -107,6 +108,7 @@ class CaliperWorker {
         await CaliperWorker._waitForTxsToFinish(stats);
     }
 
+
     /**
      * Perform test with specified test duration
      * @param {object} workloadModule The user test module.
@@ -114,11 +116,11 @@ class CaliperWorker {
      * @param {Object} rateController rate controller object
      * @async
      */
-    async runDuration(workloadModule, duration, rateController) {
+    async _runDuration(workloadModule, duration, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
         let startTime = stats.getRoundStartTime();
         let error = undefined;
-        while ((Date.now() - startTime) < (duration * 1000) && !error) {
+        while ((Date.now() - startTime) < (duration * 1000)  && !error) {
             await rateController.applyRateControl();
 
             // If this function calls this.workloadModule.submitTransaction() too quickly, micro task queue will be filled with unexecuted promises,
@@ -161,7 +163,7 @@ class CaliperWorker {
             await this.workloadModule.initializeWorkloadModule(this.workerIndex, prepareTestMessage.getWorkersNumber(), roundIndex, prepareTestMessage.getWorkloadSpec().arguments, this.connector, context);
             await CaliperUtils.sleep(this.txUpdateTime);
         } catch (err) {
-            Logger.info(`Worker [${this.workerIndex}] encountered an error during prepare test phase for round ${roundIndex}: ${(err.stack ? err.stack : err)}`);
+            Logger.warn(`Worker [${this.workerIndex}] encountered an error during prepare test phase for round ${roundIndex}: ${(err.stack ? err.stack : err)}`);
             throw err;
         } finally {
             await this.connector.releaseContext(context);
@@ -201,12 +203,11 @@ class CaliperWorker {
 
             if (testMessage.getRoundDuration()) {
                 const duration = testMessage.getRoundDuration(); // duration in seconds
-                await this.runDuration(this.workloadModule, duration, rateController);
+                await this._runDuration(this.workloadModule, duration, rateController);
             } else {
                 const number = testMessage.getNumberOfTxs();
-                await this.runFixedNumber(this.workloadModule, number, rateController);
+                await this._runFixedNumber(this.workloadModule, number, rateController);
             }
-
             Logger.debug(`Worker #${this.workerIndex} finished round #${roundIndex}`, this.internalTxObserver.getCurrentStatistics().getCumulativeTxStatistics());
             return this.internalTxObserver.getCurrentStatistics();
         } catch (err) {
--- a/packages/caliper-core/test/worker/caliper-worker.js
+++ b/packages/caliper-core/test/worker/caliper-worker.js
@@ -35,6 +35,7 @@ mockStats.getTotalSubmittedTx.onFirstCall().returns(0);
 mockStats.getTotalSubmittedTx.onSecondCall().returns(1);
 const deactivateMethod = sinon.stub();
 let logwarningMethod = sinon.stub();
+let logerrorMethod =  sinon.stub();
 
 class MockCaliperUtils {
     static resolvePath(path) {
@@ -54,7 +55,7 @@ class MockCaliperUtils {
     static getLogger() {
         return {
             debug: sinon.stub(),
-            error: sinon.stub(),
+            error: logerrorMethod,
             warn: logwarningMethod,
             info: sinon.stub()
         };
@@ -72,6 +73,8 @@ class MockInternalTxObserver {
 class MockTxObserverDispatch {
     activate() {}
 }
+
+
 MockTxObserverDispatch.prototype.deactivate = deactivateMethod;
 
 mockery.enable({
@@ -84,13 +87,14 @@ mockery.registerMock('./tx-observers/tx-observer-dispatch', MockTxObserverDispat
 
 const loggerSandbox = sinon.createSandbox();
 const CaliperUtils = require('../../lib/common/utils/caliper-utils');
-loggerSandbox.replace(CaliperUtils, "getLogger", MockCaliperUtils.getLogger);
+loggerSandbox.replace(CaliperUtils, 'getLogger', MockCaliperUtils.getLogger);
 
 const CaliperWorker = require('../../lib/worker/caliper-worker');
 
 describe('Caliper worker', () => {
     after(() => {
         loggerSandbox.restore();
+        sinon.reset();
     });
 
     describe('When executing a round', () => {
@@ -98,19 +102,13 @@ describe('Caliper worker', () => {
         const sandbox = sinon.createSandbox();
 
         beforeEach(() => {
-            logwarningMethod.reset();
-            mockRate.end.reset();
-            mockWorkload.cleanupWorkloadModule.reset();
-            mockWorkload.submitTransaction.reset();
-            mockStats.getTotalSubmittedTx.resetHistory();
-            deactivateMethod.reset();
-
             mockConnector = sinon.createStubInstance(ConnectorInterface);
             mockConnector.getContext.resolves(1);
             mockMessenger = sinon.createStubInstance(MessengerInterface);
             mockTestMessage = sinon.createStubInstance(TestMessage);
             mockTestMessage.getRateControlSpec.returns({type: '1zero-rate'});
             mockTestMessage.getWorkloadSpec.returns({module: 'test/workload'});
+            mockWorkload.initializeWorkloadModule.resolves()
             mockTestMessage.getNumberOfTxs.returns(1);
             sandbox.replace(CaliperUtils, 'resolvePath', MockCaliperUtils.resolvePath);
             sandbox.replace(CaliperUtils, 'loadModuleFunction', MockCaliperUtils.loadModuleFunction);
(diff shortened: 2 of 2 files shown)


packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -39,6 +39,7 @@ class CaliperWorker {
         this.connector = connector;
         this.workerIndex = workerIndex;
         this.messenger = messenger;
+        this.rateControlClass = RateControl;
```
davidkel, line 42: We don't need this change.

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -84,55 +85,56 @@ class CaliperWorker {
...
         const stats = this.internalTxObserver.getCurrentStatistics();
         let startTime = stats.getRoundStartTime();
-        let error = undefined;
-        while ((Date.now() - startTime) < (duration * 1000) && !error) {
+        const errors = [];
+
+        while ((Date.now() - startTime) < (duration * 1000)) {
             await rateController.applyRateControl();
 
-            // If this function calls this.workloadModule.submitTransaction() too quickly, micro task queue will be filled with unexecuted promises,
```
davidkel, line 124: Please don't delete this comment. It's important to explain why the code was implemented the way it has been.

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -84,55 +85,56 @@ class CaliperWorker {
      * @param {Object} rateController rate controller object
      * @async
      */
-    async runFixedNumber(workloadModule, number, rateController) {
+    async _runFixedNumber(workloadModule, number, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
-        let error = undefined;
-        while (stats.getTotalSubmittedTx() < number && !error) {
+        const errors = [];
```
davidkel, line 90: Sorry my mistake, it looks like the worker terminates it's run at the first error. Apologies I missed that. It means we don't need to make changes around how errors are handled

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -84,55 +85,56 @@ class CaliperWorker {
...
      * @param {Object} rateController rate controller object
      * @async
      */
-    async runDuration(workloadModule, duration, rateController) {
+    async _runDuration(workloadModule, duration, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
         let startTime = stats.getRoundStartTime();
-        let error = undefined;
-        while ((Date.now() - startTime) < (duration * 1000) && !error) {
+        const errors = [];
```
davidkel, line 122: Sorry my mistake, it looks like the worker terminates it's run at the first error. Apologies I missed that. It means we don't need to make changes around how errors are handled

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -194,19 +196,18 @@ class CaliperWorker {
 
             // Configure
             Logger.debug(`Worker #${this.workerIndex} creating rate controller`);
-            rateController = new RateControl(testMessage, this.internalTxObserver.getCurrentStatistics(), this.workerIndex);
+            rateController = new this.rateControlClass(testMessage, this.internalTxObserver.getCurrentStatistics(), this.workerIndex);
```
davidkel, line 199: We don't need this change

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -84,55 +85,56 @@ class CaliperWorker {
...
+    async _runFixedNumber(workloadModule, number, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
-        let error = undefined;
-        while (stats.getTotalSubmittedTx() < number && !error) {
+        const errors = [];
+
+        while (stats.getTotalSubmittedTx() < number) {
             await rateController.applyRateControl();
 
-            // If this function calls this.workloadModule.submitTransaction() too quickly, micro task queue will be filled with unexecuted promises,
```
davidkel, line 93: Please don't delete this comment, it's important

davidkel (changes requested):
Apologies I missed the fact that the worker stops on the first error so we don't need to make the changes to the error handling

packages/caliper-core/lib/worker/caliper-worker.js
tunedev: "I believe this can be made configurable with a flag—perhaps called **"AbortEarly"**—so that it defaults to the previous behavior. I think this approach also has the benefit of providing an overview of all errors during the initial run. What do you think?"

packages/caliper-core/lib/worker/caliper-worker.js
tunedev: i added this so i can manipulate the rate control in the test cases, what other way would you suggest i achieve this?

davidkel: Have a look a the RecordRate test. It shows a way to manipulate an existing rate control with your own mock one

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -84,55 +85,58 @@ class CaliperWorker {
...
+    async _runFixedNumber(workloadModule, number, rateController) {
         const stats = this.internalTxObserver.getCurrentStatistics();
-        let error = undefined;
-        while (stats.getTotalSubmittedTx() < number && !error) {
+        const errors = [];
+
+        while (stats.getTotalSubmittedTx() < number) {
             await rateController.applyRateControl();
 
             // If this function calls this.workloadModule.submitTransaction() too quickly, micro task queue will be filled with unexecuted promises,
```
davidkel, line 95: This comment is still missing 2 more lines, please re-instate them

packages/caliper-core/lib/worker/caliper-worker.js
davidkel: It's a possibility, but I wonder where the control for this should be here. Unfortunately there is no documented behaviour for what submitTransaction will do if it throws an error so we have to assume that the current behaviour should be that the workload module decides whether to "Abort Early" or not given that it can catch the errors and throw one if neccesary to abort. All the examples will just let errors from the connector bubble up as they are very simple examples so nothing demonstrated there either.  So technically the ability to "Abort Early" or not already exists and can be controlled in the workload module so I'm not sure we need a flag as well. Sorry I'm not the author of this code and it's not an area I am familiar with as it existed before I started working on Caliper, so I would think the best person to advise on this is Attila so I would say you should raise an issue on github with this suggestion and then let Attila and I know it exists so that we can all discuss this properly. Sorry, I don't think it should be part of this PR so this PR should not be changing the behaviour so we need to revert this change.

davidkel (changes requested):
Still some issues I'm afraid

packages/caliper-core/lib/worker/caliper-worker.js
tunedev: True that sounds like a completely different scope

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +190,101 @@ describe('Caliper worker', () => {
             await worker.executeRound(mockTestMessage);
             validateCallsAndWarnings(4);
         });
+
+        [5, 10].forEach(numberOfTxs => {
+            it(`should run ${numberOfTxs} transactions and wait for completion when no errors occur`, async () => {
```
davidkel, line 195: Need tests for when an error does occur, both for duration and fixed number runs

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +190,101 @@ describe('Caliper worker', () => {
...
+
+                sinon.assert.callCount(mockWorkload.submitTransaction, numberOfTxs);
+                sinon.assert.calledOnce(deactivateMethod);
+                sinon.assert.calledOnce(mockRate.end);
+                sinon.assert.calledOnce(mockWorkload.cleanupWorkloadModule);
+                sinon.assert.called(mockConnector.releaseContext);
+            });
+        });
+
+        it('should execute the round for a specified duration', async function() {
```
davidkel, line 233: tricky, the test doesn't prove that anything happened during this run and also doesn't prove that it took about 1 second. So you need to show that this test did do something for about the time period specified.

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +190,101 @@ describe('Caliper worker', () => {
...
+            const clock = sinon.useFakeTimers();
+            mockWorkload.submitTransaction.resolves();
+
+            mockTestMessage.getRoundDuration.returns(1);
+            mockTestMessage.getNumberOfTxs.returns(null);
+
+            const executePromise = worker.executeRound(mockTestMessage);
+
+            await clock.tickAsync(1000); // Advance time by 1 second
+            await Promise.resolve();
```
davidkel, line 278: not sure what this line is doing here could you enlighten me please ?

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +190,101 @@ describe('Caliper worker', () => {
...
+            await worker.executeRound(mockTestMessage);
+
+            sinon.assert.calledOnce(deactivateMethod);
+            sinon.assert.calledOnce(mockRate.end);
+            sinon.assert.calledOnce(mockWorkload.cleanupWorkloadModule);
+            sinon.assert.called(mockConnector.releaseContext);
+        });
+
+
+        it('should handle errors during the prepareTest phase', async () => {
```
davidkel, line 249: We need more tests for when a prepare message is received

davidkel (changes requested):
A few minor things plus ideally we could use some more tests but at least this is more than before

packages/caliper-core/test/worker/caliper-worker.js
tunedev: After advancing the virtual clock forward with `clock.tickAsync`, which triggers only the macro-tasks, I needed to ensure that all microtasks also get processed. I did this by using `await Promise.resolve();` to yield control back to the event loop.

packages/caliper-core/test/worker/caliper-worker.js
tunedev: what do you think about the update?

packages/caliper-core/test/worker/caliper-worker.js
tunedev: I added some more test specifically testing around the prepare message makeup, is this what you meant?

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +167,302 @@ describe('Caliper worker', () => {
...
+                sinon.assert.callCount(mockWorkload.submitTransaction, numberOfTxs);
+                sinon.assert.calledOnce(deactivateMethod);
+                sinon.assert.calledOnce(mockRate.end);
+                sinon.assert.calledOnce(mockWorkload.cleanupWorkloadModule);
+                sinon.assert.called(mockConnector.releaseContext);
+            });
+        });
+
+        [5, 10].forEach(numberOfTxs => {
+            it(`should handle errors while running ${numberOfTxs} transactions`, async () => {
```
davidkel, line 211: should handle errors doesn't tell me as a reader what should happen when the error occurs. Please be explicit about what happens when an error occurs. In this case the round is stopped.

packages/caliper-core/test/worker/caliper-worker.js
davidkel: Be good to add a comment here explaining that.

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -107,26 +108,31 @@ class CaliperWorker {
...
         const stats = this.internalTxObserver.getCurrentStatistics();
         let startTime = stats.getRoundStartTime();
         let error = undefined;
-        while ((Date.now() - startTime) < (duration * 1000) && !error) {
+        while ((Date.now() - startTime) < (duration * 1000)  && !error) {
+            // eslint-disable-next-line no-unused-vars
+            const now = Date.now();
             await rateController.applyRateControl();
 
+
```
davidkel, line 128: while changes are still required, be good to remove this added blank line

packages/caliper-core/lib/worker/caliper-worker.js
```diff
@@ -107,26 +108,31 @@ class CaliperWorker {
...
             // If this function calls this.workloadModule.submitTransaction() too quickly, micro task queue will be filled with unexecuted promises,
             // and I/O task(s) will get no chance to be execute and fall into starvation, for more detail info please visit:
             // https://snyk.io/blog/nodejs-how-even-quick-async-functions-can-block-the-event-loop-starve-io/
             await this.setImmediatePromise(() => {
                 workloadModule.submitTransaction()
                     .catch(err => { error = err; });
+
```
davidkel, line 135: while changes are still required, be good to remove this added blank line

packages/caliper-core/test/worker/caliper-worker.js
```diff
@@ -161,5 +167,302 @@ describe('Caliper worker', () => {
...
+
+            const callCount = mockWorkload.submitTransaction.callCount;
+            callCount.should.be.greaterThan(0);
+
+            sinon.assert.calledOnce(deactivateMethod);
+            sinon.assert.calledOnce(mockWorkload.cleanupWorkloadModule);
+            sinon.assert.called(mockConnector.releaseContext);
+        });
+
+        it('should handle errors during a duration-based round', async function() {
```
davidkel, line 285: Be more explicit here about what should happen when an error occurs. These unit tests are basically the documentation of the internal behaviour of the code base.

packages/caliper-core/test/worker/caliper-worker.js
davidkel: I think it's covered here now, although the it strings could be a little more descriptive I think you have covered what is needed.

packages/caliper-core/test/worker/caliper-worker.js
davidkel: yep, I think that's worth a try. Need to see how it behaves going forward to see if it ever fails the build being outside of the range.

davidkel (changes requested):
some final minor comments.

davidkel (approved):
Thanks @tunedev

PR #1627, opened 2024-09-11
add more test for recordRate rate-controller

tunedev:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Partially addresses #1606

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Validation of the fix

- caliper-core/lib/worker/recordRate.js: 95.52

- caliper-core/lib/worker/recordRate.js: 43.28

### DIFF
--- a/packages/caliper-core/lib/worker/rate-control/recordRate.js
+++ b/packages/caliper-core/lib/worker/rate-control/recordRate.js
@@ -47,8 +47,9 @@ class RecordRateController extends RateInterface {
 
         this.records = [];
         // if we know the number of transactions beforehand, pre-allocate the array
-        if (testMessage.getNumberOfTxs()) {
-            this.records = new Array(testMessage.getNumberOfTxs());
+        const numTx = testMessage.getNumberOfTxs();
+        if (numTx) {
+            this.records = new Array(numTx + 1);
             this.records.fill(0);
         }
 
@@ -89,7 +90,10 @@ class RecordRateController extends RateInterface {
      */
     _exportToText() {
         fs.writeFileSync(this.pathTemplate, '', 'utf-8');
-        this.records.forEach(submitTime => fs.appendFileSync(this.pathTemplate, `${submitTime}\n`));
+        for (let i = 0; i < this.records.length; i++) {
+            const time = this.records[i] !== undefined ? this.records[i] : 0;
+            fs.appendFileSync(this.pathTemplate, `${time}\n`);
+        }
     }
 
     /**
@@ -103,7 +107,8 @@ class RecordRateController extends RateInterface {
         offset = buffer.writeUInt32LE(this.records.length, offset);
 
         for (let i = 0; i < this.records.length; i++) {
-            offset = buffer.writeUInt32LE(this.records[i], offset);
+            const time = this.records[i] !== undefined ? this.records[i] : 0;
+            offset = buffer.writeUInt32LE(time, offset);
         }
 
         fs.writeFileSync(this.pathTemplate, buffer, 'binary');
@@ -120,7 +125,8 @@ class RecordRateController extends RateInterface {
         offset = buffer.writeUInt32BE(this.records.length, offset);
 
         for (let i = 0; i < this.records.length; i++) {
-            offset = buffer.writeUInt32BE(this.records[i], offset);
+            const time = this.records[i] !== undefined ? this.records[i] : 0;
+            offset = buffer.writeUInt32BE(time, offset);
         }
 
         fs.writeFileSync(this.pathTemplate, buffer, 'binary');
--- a/packages/caliper-core/test/worker/rate-control/recordRate.js
+++ b/packages/caliper-core/test/worker/rate-control/recordRate.js
@@ -17,15 +17,20 @@
 const mockery = require('mockery');
 const path = require('path');
 const RecordRate = require('../../../lib/worker/rate-control/recordRate');
+const fs = require('fs');
 const TestMessage = require('../../../lib/common/messages/testMessage');
 const MockRate = require('./mockRate');
 const TransactionStatisticsCollector = require('../../../lib/common/core/transaction-statistics-collector');
+const util = require('../../../lib/common/utils/caliper-utils');
 
-const chai = require('chai');
-chai.should();
+const { expect } = require('chai');
 const sinon = require('sinon');
 
 describe('RecordRate controller', () => {
+    let msgContent;
+    let stubStatsCollector;
+    let sandbox;
+
     before(() => {
         mockery.enable({
             warnOnReplace: false,
@@ -34,25 +39,29 @@ describe('RecordRate controller', () => {
         });
 
         mockery.registerMock(path.join(__dirname, '../../../lib/worker/rate-control/noRate.js'), MockRate);
+        sandbox = sinon.createSandbox();
     });
 
     after(() => {
         mockery.deregisterAll();
         mockery.disable();
+        if (fs.existsSync('../tx_records_client0_round0.txt')) {
+            fs.unlinkSync('../tx_records_client0_round0.txt');
+        }
     });
 
-    it('should apply rate control to the recorded rate controller', async () => {
-        const msgContent = {
+    beforeEach(() => {
+        msgContent = {
             label: 'test',
             rateControl: {
-                "type": "record-rate",
-                "opts": {
-                    "rateController": {
-                        "type": "zero-rate"
+                type: 'record-rate',
+                opts: {
+                    rateController: {
+                        type: 'zero-rate'
                     },
-                    "pathTemplate": "../tx_records_client<C>_round<R>.txt",
-                    "outputFormat": "TEXT",
-                    "logEnd": true
+                    pathTemplate: '../tx_records_client<C>_round<R>.txt',
+                    outputFormat: 'TEXT',
+                    logEnd: true
                 }
             },
             workload: {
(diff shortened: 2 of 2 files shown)


packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+
+        it('should export records to binary little endian format', async () => {
+            const msgContent = { ...defaultMsgContent };
+            msgContent.rateControl.opts.outputFormat = 'BIN_LE';
+            const testMessage = new TestMessage('test', [], msgContent);
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+            sandbox.stub(controller.recordedRateController, 'end').resolves();
+
+            controller.records = [100, 200, 300];
```
davidkel, line 174: Same comment here regarding the record data given in the text test

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            sinon.assert.calledOnce(fsWriteSyncStub);
+            const buffer = fsWriteSyncStub.getCall(0).args[1];
+            buffer.readUInt32LE(0).should.equal(3);
+            buffer.readUInt32LE(4).should.equal(100);
+            buffer.readUInt32LE(8).should.equal(200);
+            buffer.readUInt32LE(12).should.equal(300);
+
+            fsWriteSyncStub.restore();
+        });
+        it('should export to text format when output format is TEXT', async () => {
```
davidkel, line 188: This is already tested above. given that you track the writes to the file, we know it's only writing to text and not the others.

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            sinon.assert.notCalled(exportToBinaryLittleEndianSpy);
+            sinon.assert.notCalled(exportToBinaryBigEndianSpy);
+            sinon.assert.notCalled(logger.error);
+
+            exportToTextSpy.restore();
+            exportToBinaryLittleEndianSpy.restore();
+            exportToBinaryBigEndianSpy.restore();
+        });
+
+        it('should export to binary little endian format when output format is BIN_LE', async () => {
```
davidkel, line 215: I don't think we need this test as it's already tested.

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            sinon.assert.notCalled(exportToBinaryBigEndianSpy);
+            sinon.assert.notCalled(logger.error);
+
+            exportToTextSpy.restore();
+            exportToBinaryLittleEndianSpy.restore();
+            exportToBinaryBigEndianSpy.restore();
+        });
+
+
+        it('should export to binary big endian format when output format is BIN_BE', async () => {
```
davidkel, line 244: I don't think we need this test as it's already tested

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            msgContent.rateControl.opts.pathTemplate = undefined;
+            const testMessage = new TestMessage('test', [], msgContent);
+
+            (() => {
+                RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+            }).should.throw('The path to save the recording to is undefined');
+        });
+    });
+
+    describe('Rate Control', () => {
```
davidkel, line 283: This describe could be changed to "When Applying Rate Control"

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            const testMessage = new TestMessage('test', [], defaultMsgContent);
+            const rateController = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+            const mockRate = MockRate.createRateController();
+            mockRate.reset();
+            mockRate.isApplyRateControlCalled().should.equal(false);
+            await rateController.applyRateControl();
+            mockRate.isApplyRateControlCalled().should.equal(true);
+        });
+
+        it('should initialize records array if the number of transactions is provided', () => {
```
davidkel, line 294: We need a separate describe here of "When Creating a RecordRate Controller"

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            const msgContent = {...defaultMsgContent };
+
+            const testMessage = new TestMessage('test', [], msgContent);
+            sinon.stub(testMessage, 'getNumberOfTxs').returns(5);
+
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+            controller.records.should.be.an('array').that.has.lengthOf(5);
```
davidkel, line 302: Here you have written a test that passes when it fact there is a bug here. The length should be 6, because the total possible values for the number of transactions submitted is 0-5.

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            controller.records.every(record => record.should.equal(0));
+        });
+
+        it('should replace path template placeholders', () => {
+            const msgContent = { ...defaultMsgContent };
+            const testMessage = new TestMessage('test', [], msgContent);
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+            controller.pathTemplate.should.equal(util.resolvePath('../tx_records_client0_round0.txt'));
```
davidkel, line 311: would be good to include more than just 0 here. We should also test with different worker index and round numbers. What if the bug was that it's has some hardcoded value of 0 in the code somewhere ? This test would pass and not discover the bug

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +74,262 @@ describe('RecordRate controller', () => {
...
+            msgContent.rateControl.opts.logEnd = true;
+            const testMessage = new TestMessage('test', [], msgContent);
+
+            (() => {
+                RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+            }).should.throw(/Module "nonexistent-rate" could not be loaded/);
+        });
+
+        it('should throw an error if rateController is undefined', () => {
+            const msgContent = { ...defaultMsgContent };
```
davidkel, line 326: same shallow copy comment applies here

davidkel (changes requested):
There appear to be more changes required for this based on previous comments.

tunedev:
> There appear to be more changes required for this based on previous comments.

@davidkel i have tended to them, thanks

packages/caliper-core/lib/worker/rate-control/recordRate.js
```diff
@@ -103,7 +106,8 @@ class RecordRateController extends RateInterface {
         offset = buffer.writeUInt32LE(this.records.length, offset);
 
         for (let i = 0; i < this.records.length; i++) {
-            offset = buffer.writeUInt32LE(this.records[i], offset);
+            const time = this.records[i] !== undefined ? this.records[i] : 0;
```
davidkel, line 109: technically these changes are not required because an empty element in an array will still be written out as zero. However I have no problem leaving it in as it means we are more explicit in declaring what we want the behaviour to be rather than relying on what the writeUInt32LE method does. But you didn't add it for the writing of BE as well. So either we have it in both places, or not in either place

packages/caliper-core/lib/worker/rate-control/recordRate.js
```diff
@@ -132,7 +136,7 @@ class RecordRateController extends RateInterface {
      */
     async applyRateControl() {
         await this.recordedRateController.applyRateControl();
-        this.records[this.stats.getTotalSubmittedTx()] = Date.now() - this.stats.getRoundStartTime();
+        this.records[this.stats.getTotalSubmittedTx() - 1] = Date.now() - this.stats.getRoundStartTime();
```
davidkel, line 139: Definitely not add the `-1`. I don't know why you did that ?

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +73,209 @@ describe('RecordRate controller', () => {
...
+    });
+
+    describe('When Creating a RecordRate Controller', () => {
+        it('should initialize records array if the number of transactions is provided', () => {
+            const testMessage = new TestMessage('test', [], msgContent);
+            sinon.stub(testMessage, 'getNumberOfTxs').returns(5);
+
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+            controller.records.should.be.an('array').that.has.lengthOf(5);
```
davidkel, line 275: this test is not correct, as stated in a previous review, the value here should be one more than the value returned by getNumberOfTxs because we can and probably will have an entry for 0 transactions submitted therefore if we have 5 transactions submitted then we would have 0-5 possible records. This will then show there is a bug in the code which needs to be fixed.

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +73,209 @@ describe('RecordRate controller', () => {
...
+            stubStatsCollector.getTotalSubmittedTx.returns(1);
```
davidkel, line 276: This line doesn't do anything of use, so can be removed

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +73,209 @@ describe('RecordRate controller', () => {
...
+            controller.records.every(record => record.should.equal(0));
```
davidkel, line 277: Ideally I would prefer the use of expect here purely because of for some reason record was set to null or undefined, the test fails with a invoke of a method on a non existent object, which is why I prefer expect over should now. But not a big deal as the test still fails.

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +73,209 @@ describe('RecordRate controller', () => {
...
+                testRound: 0,
+                txDuration: 250,
+                totalWorkers: 2
+            };
+            const testMessage = new TestMessage('test', [], msgContent);
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+            sinon.stub(controller.recordedRateController, 'end').resolves();
+
+            // Use non-sequential indices with gaps
+            controller.records = new Array(8).fill(undefined); // Initialize with length 8
```
davidkel, line 151: As explained above, please remove this line. It won't break the test

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +73,209 @@ describe('RecordRate controller', () => {
...
+
+        it('should export records to binary little endian format with gaps', async () => {
+            msgContent.rateControl.opts.outputFormat = 'BIN_LE';
+            const testMessage = new TestMessage('test', [], msgContent);
+            const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+            sandbox.stub(controller.recordedRateController, 'end').resolves();
+
+            // Create an array with gaps to simulate non-sequential transaction submissions
+            controller.records = new Array(8).fill(undefined); // Initialize with length 8
```
davidkel, line 185: As explained above, please remove this line. It won't break the test

davidkel (changes requested):
So more issues here I'm afraid, but the sparse population of the records array has shown up  bugs but unfortunately you hid by fixing the test, so something to always bare in mind.
On further thought as well for completeness we should also add tests that test records that are not sparse. ie we have continuous values starting at the 0'th record element (eg records[0]=100;records[1]=300;records[2]=500;records[3]=700;) as this is a common scenario as well. So we should have tests for it.

packages/caliper-core/lib/worker/rate-control/recordRate.js
```diff
@@ -89,7 +89,10 @@ class RecordRateController extends RateInterface {
      */
     _exportToText() {
         fs.writeFileSync(this.pathTemplate, '', 'utf-8');
-        this.records.forEach(submitTime => fs.appendFileSync(this.pathTemplate, `${submitTime}\n`));
+        this.records.forEach(submitTime => {
```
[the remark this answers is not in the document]
tunedev: Oh okay, would look at the docs to fix the bug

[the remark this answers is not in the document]
davidkel: @tunedev it's something that's been in the code for a while by the looks of things and it works when only one request is sent in the sendRequests call, but now thanks to you putting tests in which would simulate multiple requests in a sendRequests call, it would fail to do the right thing.

packages/caliper-core/test/worker/rate-control/recordRate.js
davidkel: Correct, we want to pre-allocate an array that can cope with 0 to T elements where T is the number of transactions, therefore we need the array allocate T+1 elements. try it for yourself:
```
let numTx = 5;
let x = new Array(numTx);
console.log(x.length) // returns 5
x[numTx]=1;
console.log(x.length) // returns 6
```

tunedev: oh yeah it is obvious now, i was focussed on guarding the existing implementation, that i lost track of how it was meant to be used

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +72,171 @@ describe('RecordRate controller', () => {
...
+            recordScenarios.forEach(scenario => {
+                it(`should export records to ${format} format ${scenario.description}`, async () => {
+                // Prepare message content with the specific output format
+                    const msgContentCopy = JSON.parse(JSON.stringify(msgContent));
+                    msgContentCopy.rateControl.opts.outputFormat = format;
+                    const testMessage = new TestMessage('test', [], msgContentCopy);
+                    const controller = RecordRate.createRateController(testMessage, stubStatsCollector, 0);
+
+                    sinon.stub(controller.recordedRateController, 'end').resolves();
+                    controller.records = new Array(scenario.expectedLength);
```
davidkel, line 124: we should remove this line, the test should not be setting up the controller records array, that is the responsibility of the rate controller constructor

packages/caliper-core/test/worker/rate-control/recordRate.js
```diff
@@ -63,42 +72,171 @@ describe('RecordRate controller', () => {
...
+                    });
+
+                    const fsWriteSyncStub = sandbox.stub(fs, 'writeFileSync');
+                    const fsAppendSyncStub = sandbox.stub(fs, 'appendFileSync');
+
+                    await controller.end();
+
+                    if (format === 'TEXT') {
+                        sinon.assert.calledOnce(fsWriteSyncStub);
+                        sinon.assert.callCount(fsAppendSyncStub, controller.records.length);
```
davidkel, line 137: this should assert to expectedLength and we should also assert that the records.length asserts to the expectedLength

davidkel (changes requested):
A few more changes please

davidkel (approved):
LGTM Thanks

PR #1628, opened 2024-09-11
add more test for fixedFeedbackRate controller implementation

tunedev:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Partially addresses #1606

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Validation of the fix

- caliper-core/lib/worker/fixedFeedbackRate.js: 80

- caliper-core/lib/worker/fixedFeedbackRate.js: 60

### DIFF
--- a/packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
+++ b/packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
@@ -78,7 +78,7 @@ describe('fixedFeedbackRate controller implementation', () => {
     });
 
     describe('#applyRateController', () => {
-        let controller, sleepStub, txnStats, clock;
+        let controller, sleepStub, txnStats, clock, testMessage;
 
         beforeEach( () => {
             const msgContent = {
@@ -99,7 +99,7 @@ describe('fixedFeedbackRate controller implementation', () => {
             sleepStub = sinon.stub();
             FixedFeedbackRate.__set__('util.sleep', sleepStub);
 
-            const testMessage = new TestMessage('test', [], msgContent);
+            testMessage = new TestMessage('test', [], msgContent);
             txnStats = new TransactionStatisticsCollector();
             controller = new FixedFeedbackRate.createRateController(testMessage, txnStats, 0);
         });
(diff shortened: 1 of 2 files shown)


packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -144,5 +144,29 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+
+            sinon.assert.calledOnce(sleepStub);
+            sinon.assert.calledWith(sleepStub, 1 * controller.backOffTime);
+        });
+
+        it('should sleep when unfinished transactions are above the threshold', async () => {
+            txnStats.stats.txCounters.totalSubmitted = 44;
+            txnStats.stats.txCounters.totalFinished = 40;
+            controller.unfinishedPerWorker = 1;
+            controller.backOffTime = 100;
```
davidkel, line 164: this line is redundent as it's the default value

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -144,5 +144,29 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+            await controller.applyRateControl();
+
+            sinon.assert.calledOnce(sleepStub);
+            sinon.assert.calledWith(sleepStub, 1 * controller.backOffTime);
+        });
+
+        it('should sleep when unfinished transactions are above the threshold', async () => {
+            txnStats.stats.txCounters.totalSubmitted = 44;
+            txnStats.stats.txCounters.totalFinished = 40;
+            controller.unfinishedPerWorker = 1;
```
davidkel, line 163: This line is also redundent and will cover it later, however usually I would prefer that this value be defined via the construction of the rate control as the test should be about what a specifically configured rate controller does under different transaction completion state conditions

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -144,5 +144,29 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+            controller.backOffTime = 100;
+
+            await controller.applyRateControl();
+
+            sinon.assert.calledOnce(sleepStub);
+            sinon.assert.calledWith(sleepStub, 88 * controller.backOffTime);
```
davidkel, line 169: this line, although the value is correct for the number sent to the sleep stub, it implies the backofTime is relevant to how the sleep value was calculated and thus the code path. It's in fact calculated as follows 
        let diff = (this.generalSleepTime * currentSubmitted - ((Date.now() - this.totalSleepTime) - this.stats.getRoundStartTime()));
and this is actually 44*200 = 8800

See also the comment about the description for further details

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -144,5 +144,29 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+            txnStats.stats.txCounters.totalSuccessful = 0;
+            controller.generalSleepTime = 1;
+
+            await controller.applyRateControl();
+
+            sinon.assert.calledOnce(sleepStub);
+            sinon.assert.calledWith(sleepStub, 1 * controller.backOffTime);
+        });
+
+        it('should sleep when unfinished transactions are above the threshold', async () => {
```
davidkel, line 160: Unfortunately this doesn't tell me exactly what is being tested here. I know it's about unfinished transactions but have no idea what you mean by the "threshold". So as someone who wants to look at the tests to understand the behaviour using terms that are not defined as part of the configuration of the rate controller is only going to confuse. 

But this also brings me to the code around this area which is lacking. I can only assume you weren't trying to test the following code based on the test
```
        let diff = (this.generalSleepTime * currentSubmitted - ((Date.now() - this.totalSleepTime) - this.stats.getRoundStartTime()));
        if (diff > 5) {
            await util.sleep(diff);
            return;
        }
   ```
but if you were, then the issue with the code is that the variable `diff` is no meaning whatsoever as to the value it represents. To be honest I would have to spend time thinking about it myself to understand what it actually represents and why the value 5 was chosen. This is not technically a bug, but is certainly poor programming as code needs to convey as much information as possible and where that cannot be done then you use comments. 

The point I am making here is that this description doesn't explain the test to me because of the use of the word threshold.

**Review by davidkel — CHANGES_REQUESTED** [review]
Sorry, this is going to be a hard one to test and make sure that the descriptions of the tests provide detail. Unfortunately the original tests are not ideal either but I don't think we should follow the same pattern as this file if we know they are not ideal. If possible it would be nice to update them but I think that would have to be an exercise for later.

However I saw you correctly used async/await in your tests and the original ones didn't, so a quick update worth doing would be to apply that pattern to the original tests.

File: packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js [hunk 5]
```diff
@@ -108,41 +108,190 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+            await controller.applyRateControl();
 
             // should have not called the sleep method
             sinon.assert.notCalled(sleepStub);
         });
 
-        it('should not sleep if there are no unfinished transactions', () => {
-            txnStats.stats.txCounters.totalSubmitted = 0;
-            controller.applyRateControl();
+        // it('should not sleep if there are no unfinished transactions', async () => {
```
davidkel, line 120: Test is commented out

davidkel (changes requested):
You've commented out a test

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
tunedev: Ooh forgot to clean it up, fixed

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -108,41 +108,176 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+                clock.tick(12);
+
+                await controller.applyRateControl();
+
+                const expectedSleepTime = i * controller.backOffTime;
+
+                sinon.assert.calledOnce(sleepStub);
+                sinon.assert.calledWith(sleepStub, expectedSleepTime);
+
+                // Restore stubs
```
davidkel, line 275: Why do this, txnStats is recreated on every test as it's in the beforeEach

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -108,41 +108,176 @@ describe('fixedFeedbackRate controller implementation', () => {
...
 
             // should have not called the sleep method
             sinon.assert.notCalled(sleepStub);
         });
 
-        it('should not sleep if there are no unfinished transactions', () => {
-            txnStats.stats.txCounters.totalSubmitted = 0;
-            controller.applyRateControl();
+        it('should not sleep if there are no unfinished transactions', async () => {
+            // Stub methods if necessary
```
davidkel, line 121: Why stub only in this test when all others, including the original test just set the stats directly ?

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -108,41 +108,176 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+
+                // Stub getRoundStartTime
+                sinon.stub(txnStats, 'getRoundStartTime').returns(0);
+
+                // Stub getTotalSubmittedTx and getTotalFinishedTx
+                const unfinished = i * unfinishedPerWorker;
+                const totalFinished = 10;
+                const totalSubmitted = totalFinished + unfinished;
+
+                sinon.stub(txnStats, 'getTotalSubmittedTx').returns(totalSubmitted);
```
davidkel, line 261: again wonder why you stub here, but all others you set the txnStats values directly

packages/caliper-core/test/worker/rate-control/fixedFeedbackRate.js
```diff
@@ -108,41 +108,176 @@ describe('fixedFeedbackRate controller implementation', () => {
...
+            txnStats.stats.txCounters.totalSuccessful = 0;
+
+            controller.zeroSuccessfulCounter = 3;
+            clock.tick(0); // Ensure clock starts at 0
+
+            await controller.applyRateControl();
+
+            sinon.assert.calledOnce(sleepStub);
+            // Expected sleep time should be 4 * backOffTime = 400ms
+            sinon.assert.calledWith(sleepStub, 4 * controller.backOffTime);
```
davidkel, line 183: This is 1 more than the zeroSuccessfulCounter, would be good to show this in this test

davidkel (approved):
I have some minor comments, but given timeframes left to complete this I think we can merge this as is.

PR #1643, opened 2024-10-19
add dev container for vscode

psankhe28:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,18 @@
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.8
+
+RUN apt-get update && apt-get install -y \
+    git \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /workspace
+
+COPY . .
+
+RUN pip install -r docs/pip-requirements.txt
+
+ENV MKDOCS_PORT=8000
+
+EXPOSE ${MKDOCS_PORT}
+
+CMD ["mkdocs", "serve", "-a", "0.0.0.0:8000"]
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,21 @@
+{
+    "name": "Caliper Development",
+    "context": "..",
+    "dockerFile": "Dockerfile",
+    "customizations": {
+        "vscode": {
+            "settings": {
+                "terminal.integrated.shell.linux": "/bin/bash",
+                "python.defaultInterpreterPath": "/usr/local/bin/python"
+            },
+            "extensions": [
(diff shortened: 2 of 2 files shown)


davidkel (changes requested):
Code you also add node support into this as well, the idea would be that this dev container has everything you need to develop caliper. ie you don't have to install anything. I suggest node 20

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.8
```
davidkel, line 1: We actually don't need this file. We can do everything we need inside the devcontainer.json

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+                "ms-python.python",
+                "esbenp.prettier-vscode",
+                "ms-azuretools.vscode-docker"
+            ]
+        }
+    },
+    "postCreateCommand": "cd docs && pip install -r pip-requirements.txt && mkdocs build",
+    "remoteUser": "vscode",
+    "forwardPorts": [8000] 
+}
```
davidkel, line 21: We will build everything we need in this file using the features capability of devcontainers. Here is my proposal for the solution. As you can see it takes what you proposed and simplifies it. It also adds in docker support which will be useful for running the integration tests (although I've not tried that yet, it could be built on if it doesn't work)
```json
{
    "name": "Caliper Development",
    "image": "mcr.microsoft.com/vscode/devcontainers/python:3.11",
    "features": {
        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
        "ghcr.io/devcontainers/features/node:1": {
            "version": "lts"
        }
    },
    "customizations": {
        "vscode": {
            "settings": {
                "terminal.integrated.shell.linux": "/bin/bash",
                "python.defaultInterpreterPath": "/usr/local/bin/python"
            },
            "extensions": [
                "esbenp.prettier-vscode"
            ]
        }
    },
    "postCreateCommand": "cd docs && pip install -r pip-requirements.txt && mkdocs build",
    "remoteUser": "vscode",
    "forwardPorts": [8000]
}
```
So this selects the image as defined in the dockerfile given and then easily adds node and docker support. The node version that will be installed here is 22 but that's ok we need to support node 22 now.
I've removed the python extension there is no python development required for caliper. I've also removed the vscode-docker extension as I'm not sure that's useful at the moment.

davidkel (changes requested):
My proposal is to remove the Dockerfile file and update devcontainer.json as proposed in the review comments.

davidkel (approved):
LGTM
Many thanks

PR #1644, opened 2024-10-22
creates build and pr to gh-page

psankhe28:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


davidkel:
@psankhe28 I don't think this is what we will go with as discussed previously. So here is my suggestions

1. Submit a PR to the gh-pages branch that replaces the documentation with the new version of docs in the main branch
2. Write a section in the release_process.md file in main about how to add another version of the docs (for example v1.0.0) to gh-pages via a pull request. The instructions should set that to the default version (but ensure that the 0.6.0 version still remains) and it would then be submitted to gh-pages as a PR manually. 

Investigating how `mike` works the whole of the gh-pages branch content will be managed by mike, therefore we can take a different approach to managing the content which still retains branch protection but allows github actions to automate publication without the need to generate PRs. This is not something that needs investigation at this time though so please don't look into it.

davidkel (changes requested):
We aren't going with a PR approach for github actions

PR #1646, opened 2024-10-29
docs: update links to point to new website URL

argbet21:


packages/caliper-ethereum/lib/ethereum-connector.js
```diff
@@ -67,7 +67,7 @@ class EthereumConnector extends ConnectorBase {
         if (!ethereumConfig.url) {
             throw new Error(
                 'No URL given to access the Ethereum SUT. Please check your network configuration. ' +
-                'Please see https://hyperledger.github.io/caliper/v0.3/ethereum-config/ for more info.'
+                'Please see https://hyperledger-caliper.github.io/caliper/v0.3/ethereum-config/ for more info.'
```
davidkel, line 70: The code has a reference to an old version of the docs ? If you could just delete this line and the plus sign on the previous line that would be great.

davidkel (commented):
You need to provide DCO sign off of your commit, plus if you could make one minor change that would be greatly appreciated. Many thanks for the contribution.

argbet21:
Alright, done!

I also noticed that the link you requested to delete actually exists in the 0.6.0 version. So instead of removing it, I updated it to point to the latest Caliper version. However, let me know if it’s still preferable to remove it altogether.

davidkel:
@argbet21 I would prefer to remove it

argbet21:
Done! Let me know if there's anything else that's ideal to change.

davidkel (approved):
LGTM thanks for the contribution

Issue #1647, opened 2024-10-29
Report logic may not be able to find the hyperledger caliper logo when new docs are deployed


davidkel:
packages/caliper-core/lib/manager/report/template/report.html has the following line
```
<img src="https://caliper.github.io/caliper/assets/img/hyperledger_caliper_logo_color.png" style="width:95%;" alt="">
```

this may not work anymore even if changed to 
```
<img src="https://hyperledger-caliper.github.io/caliper/assets/img/hyperledger_caliper_logo_color.png" style="width:95%;" alt="">
```

PR #1649, opened 2024-10-30
chore: minor adjustments

argbet21:
Resolves issue #1647.

Lmk if this change addresses the issue or if you were thinking of something else.


packages/caliper-core/lib/manager/report/template/report.html
```diff
@@ -122,7 +122,7 @@
 <body style="font-family: IBM Plex Sans; font-weight: 200;">
     <main>
         <div class="left-column">
-            <img src="https://hyperledger-caliper.github.io/caliper/assets/img/hyperledger_caliper_logo_color.png" style="width:95%;" alt="">
+            <img src="/assets/hyperledger_caliper_logo_color.png" style="width:95%;" alt="">
```
davidkel, line 125: Unfortunately at the moment I don't think this would work. We need to consider how uses of caliper say via npm modules would work, plus if they publish the report somwhere they would need to be able to get the file from somewhere. Not sure what the best solution is to this at this point.

davidkel (changes requested):
Sorry, not sure this works and I haven't really thought about what would be better than referencing the file from the served docs at the moment.

Issue #1635, opened 2024-09-18
Documentation should state more clearly the rows of record rate and their meaning


davidkel:
The rows should correspond to number of submitted transactions, but it's not clear about the value of the first row. In this case as we can have 0 submitted transactions, then the first rows start from 0. This is not clear in the docs.

PR #1653, opened 2024-11-16
fix: record rate doc section

psankhe28:
## Checklist
 - [x]  A link to the issue/user story that the pull request relates to
 - [ ]  Design of the fix

## Issue/User story
#1635 

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

### DIFF
--- a/docs/source/concepts/rate-controllers.md
+++ b/docs/source/concepts/rate-controllers.md
@@ -296,23 +296,15 @@ Data:   |length |1st    |2nd    |3rd     |4th     |...
 ```
 
 - The first entry (length) indicates the number of transactions submitted.
-- Each subsequent entry in the array represents the timestamp of a transaction being enabled (submitted) by the rate controller.
+- Each subsequent entry in the array represents a timestamp of when that number of transactions were submitted from the start of the round.
 - The subsequent values (Uint32 values in Big Endian or Little Endian format) represent time in milliseconds from the start of the round. Each entry corresponds to the number of transactions submitted starting from 0
 
 For example, if there are 3 transactions enabled, the binary format would look like:
 
 ```sh
-Offset: |0      |4      |8      |12      
-Data:   |length |1st    |2nd    |3rd
-        |3      |100    |200    |300
-```
-
-If no transactions are submitted, it would look like:
-
-```sh
-Offset: |0     
-Data:   |length
-        |0
+Offset: |0      |4      |8      |12     |16      
+Data:   |length |0 tx   |1 tx   |2 tx   |3 tx 
+        |4      |100    |200    |300    |400
 ```
 
 ## Replay rate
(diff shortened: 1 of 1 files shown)


davidkel (changes requested):
Not sure this really addresses the issue here and is also confusing. Let's start with the text output format.
1. we should just state that the value of i will start at zero, so the first line represents 0 transactions submitted. Each value in a line represents  time in milliseconds from the start of the round
2. 
- The first entry length is the total number of transactions submitted (it isn't `so far`) as it's a final output
- remove the line that starts `if no transactions have been submitted.....` this is not correct
- the subsequent values (Uint32 values in Big Endian or Little Endian format) represent time in milliseconds from the start of the round. Each entry corresponds to the number of transactions submitted starting from 0

docs/source/concepts/rate-controllers.md
```diff
@@ -285,13 +285,36 @@ The template can (**and should**) contain special “variables/placeholders” t
...
+
+For example, if there are 3 transactions enabled, the binary format would look like:
+
+```sh
+Offset: |0      |4      |8      |12      
+Data:   |length |1st    |2nd    |3rd
+        |3      |100    |200    |300
+```
+
+If no transactions are submitted, it would look like:
```
davidkel, line 310: This isn't correct, so please remove the section starting "If no transactions are submitted, it would look like:"

docs/source/concepts/rate-controllers.md
```diff
@@ -285,13 +285,36 @@ The template can (**and should**) contain special “variables/placeholders” t
...
 ```sh
 Offset: |0      |4      |8      |12      |16      |...     
 Data:   |length |1st    |2nd    |3rd     |4th     |...      
 ```
 
+- The first entry (length) indicates the number of transactions submitted.
+- Each subsequent entry in the array represents the timestamp of a transaction being enabled (submitted) by the rate controller.
+- The subsequent values (Uint32 values in Big Endian or Little Endian format) represent time in milliseconds from the start of the round. Each entry corresponds to the number of transactions submitted starting from 0
+
+For example, if there are 3 transactions enabled, the binary format would look like:
```
davidkel, line 302: For the above sentence, if should say `if there are 3 transactions submitted`
The example here is wrong. If there were 3 transactions submitted then there would be 4 entries
```
Offset: |0      |4      |8      |12     |16      
Data:   |length |0 tx   |1 tx   |2 tx   |3 tx 
        |4      |100    |200    |300    |400

```

docs/source/concepts/rate-controllers.md
```diff
@@ -285,13 +285,36 @@ The template can (**and should**) contain special “variables/placeholders” t
...
+
 **Binary format**: Both binary representations encode the `X` number of recordings as a series of `X+1` UInt32 numbers (1 number for the array length, the rest for the array elements), either in Little Endian or Big Endian encoding:
 
 ```sh
 Offset: |0      |4      |8      |12      |16      |...     
 Data:   |length |1st    |2nd    |3rd     |4th     |...      
 ```
 
+- The first entry (length) indicates the number of transactions submitted.
+- Each subsequent entry in the array represents the timestamp of a transaction being enabled (submitted) by the rate controller.
```
davidkel, line 299: replace the line 
```
Each subsequent entry in the array represents the timestamp of a transaction being enabled (submitted) by the rate controller
```
with 
```
Each subsequent entry in the array represents a timestamp of when that number of transactions were submitted from the start of the round.
```

davidkel (changes requested):
Some changes required, suggestions are in the comments

davidkel (approved):
LGTM Many thanks

PR #1719, opened 2025-10-02
chore(generator)!: Remove generator package and related artifacts

aklenik:


README.md
```diff
@@ -15,7 +14,8 @@ Welcome to the Hyperledger Caliper project. Caliper is a blockchain performance
 
 Currently supported blockchain solutions:
 
-* [Hyperledger Fabric v1.X, v2.X](https://github.com/hyperledger/fabric)
+* [Hyperledger Fabric v2.X (Deprecated Network SDK, Gateway SDK)](https://github.com/hyperledger/fabric/tree/v2.5.13)
+* [Hyperledger Fabric v3.X (Gateway SDK)](https://github.com/hyperledger/fabric)
```
davidkel, line 18: I was trying to avoid claiming fabric 3.x support for a couple of reasons
1. We don't test it at all in the integration tests
2. Fabric 3.x is not LTS and I don't think whoever is left on Fabric cares about creating an LTS release (https://github.com/hyperledger/fabric/issues/5259)

aklenik: Fair enough, updated.

davidkel (approved):
Only a comment about supporting Fabric 3.x but no problem if you want to merge as is

PR #1721, opened 2025-10-03
chore: Bump to 0.8.0-unstable version

aklenik:


aklenik:
> Wouldn't 0.7.1 be better as I can't imagine any breaking changes going in for a while, but things like npm dependency changes and unit tests are the most likely changes for now

@davidkel I'm thinking about reviewing the API of different pluggable modules. Those grew mainly organically, but there should be some common life-cycle management around them. This would be a key point in separating heavy plugins with specific dependencies (like the Fabric connector, or monitors) into their own package (and repo, for security decoupling), which could lead to repo templates for the community.

So bigger changes might be coming 😄

davidkel (approved):

PR #1722, opened 2025-10-03
Fix dev container

davidkel:
Docker Outside of Docker changed something and now we have to explicitly tell it not to use Moby


aklenik (approved):
Looks good

PR #1733, opened 2025-11-13
docs: fixed typo 'an' -> 'and'

aryn-rjpt:
This commit corrects a small grammatical error in the documentation where 'an' was used instead of 'and'.


aryn-rjpt:
You're welcome :)

